PR #18, opened 2014-12-03
rebased _bulk_get patch

dholth:
This is a rebase for https://issues.apache.org/jira/browse/COUCHDB-2310 against post-1.6.x CouchDB.

The idea is that when it is finished *every* GET option should eventually be supported in the _bulk_get API, making CouchDB easier to learn than a hypothetical _bulk_get API supporting only a subtly different subset of the GET API.

Is it acceptable to refactor the GET handling?

Is anyone willing to help get this polished and accepted into CouchDB?

I wrote this patch because without it CouchDB replication to PouchDB is too slow to be useful. After this patch, formerly impractical replications that would take ~10,000 requests now take a pleasantly acceptable ~100 requests.

https://github.com/dholth/pouchdb-bulk-get is a PouchDB plugin for the other end.

Thanks.

### DIFF
--- a/src/couch_httpd_db.erl
+++ b/src/couch_httpd_db.erl
@@ -361,6 +361,61 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_bulk_docs">>]}=Req, Db) ->
 db_req(#httpd{path_parts=[_,<<"_bulk_docs">>]}=Req, _Db) ->
     send_method_not_allowed(Req, "POST");
 
+
+% Allow all options of a regular 'GET' request in a batch.
+db_req(#httpd{method='POST',path_parts=[_,<<"_bulk_get">>]}=Req, Db) ->
+    couch_stats_collector:increment({httpd, bulk_get}),
+    couch_httpd:validate_ctype(Req, "application/json"),
+    {JsonProps} = couch_httpd:json_body_obj(Req),
+    case couch_util:get_value(<<"docs">>, JsonProps) of
+    undefined ->
+        send_error(Req, 400, <<"bad_request">>, <<"Missing JSON list of 'docs'">>);
+    DocsArray ->
+            % expect an array of [{id:x, other get parameters}]
+            % for PouchDB, absolutely must support only
+            % {id: id, open_revs: missingBatch, revs: true, attachments: true}
+            % for proposed spec need to support all GET parameters
+            {ok, Resp} = start_json_response(Req, 200),
+            send_chunk(Resp, "["),
+            lists:foldl(fun(JsonObj, Separator) -> 
+                     {Props} = JsonObj,
+                     DocId = couch_util:get_value(<<"id">>, Props),
+                     couch_doc:validate_docid(DocId),
+                     #doc_query_args{
+                        rev = Rev,
+                        open_revs = Revs,
+                        options = Options1,
+                        atts_since = AttsSince
+                       } = parse_doc_args(Props),
+                    send_chunk(Resp, Separator),
+                    % todo handle case where we don't have all options
+                    {ok, Results} = couch_db:open_doc_revs(Db, DocId, Revs, Options1),
+                    send_chunk(Resp, "["),
+                    lists:foldl(
+                      fun(Result, AccSeparator) ->
+                          case Result of
+                          {ok, Doc} ->
+                              JsonDoc = couch_doc:to_json_obj(Doc, Options1),
+                              Json = ?JSON_ENCODE({[{ok, JsonDoc}]}),
+                              send_chunk(Resp, AccSeparator ++ Json);
+                          {{not_found, missing}, RevId} ->
+                              RevStr = couch_doc:rev_to_str(RevId),
+                              Json = ?JSON_ENCODE({[{"missing", RevStr}]}),
+                              send_chunk(Resp, AccSeparator ++ Json)
+                          end,
+                          "," % AccSeparator now has a comma
+                      end,
+                      "", Results),
+                      send_chunk(Resp, "]"),
+                    "," % Separator now has a comma
+                 end, "", DocsArray),
+            send_chunk(Resp, "]"),
+            end_json_response(Resp)
+    end;
+db_req(#httpd{path_parts=[_,<<"_bulk_get">>]}=Req, _Db) ->
+    send_method_not_allowed(Req, "POST");
+
+
 db_req(#httpd{method='POST',path_parts=[_,<<"_purge">>]}=Req, Db) ->
     couch_httpd:validate_ctype(Req, "application/json"),
     {IdsRevs} = couch_httpd:json_body_obj(Req),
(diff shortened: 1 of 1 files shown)


kxepal:
Hi there, short update.

I'd started porting this PR over 2.0. First note that in current state it doesn't works and breaks any document update, so don't try to merge it (:

Second, while fixing it I found that initial format doesn't assumes any possible errors:

```
$ echo '{"docs": [{"id": "foo"}, {"id": "bar"}]}' | http http://localhost:15986/test/_bulk_get  
[
    [{"_id": "foo", "_rev": "1-967a00dff5e02add41819138abb3284d"}],
    [{"_id": "bar", "_rev": "1-6aaf7080aad9d1a9482e07c46581dac0"}]
]
```

What the response should be if document id is missed or wrong or document not exists or some parameters are malformed? Oblivious solution is:

```
$ echo '{"docs": [{"id": "foo"}, {"id": "bar"}]}' | http http://localhost:15986/test/_bulk_get  
[
    [{"_id": "foo", "_rev": "1-967a00dff5e02add41819138abb3284d"}],
    [{"error": "not_found", "id": "bar", "reason": "not today"}]
]
```

But such array of arrays makes me sad panda.

I'd started to look on current implementations and found no agreement on what the response should looks like:

- [Couchbase](http://developer.couchbase.com/mobile/develop/references/sync-gateway/rest-api/database/post-bulk-get/index.html) assumes the response to be always multipart and request payload should contain array of objects where "id" field is mandatory, "rev" and "atts_since"  are optional and no mention of the others. So, passing open_revs will not work for Couchbase or make it incompatible with.

- [RCouch](https://github.com/rcouch/couchdb-couch-httpd/blob/master/src/couch_httpd_bulk_get.erl#L45) returns JSON response in the fashion `{"results": [{"id": "foo", "docs": [...]}, ...]}` what is completely different from proposed patch and pouchdb-bulk-get implementation.

- Each of these are incompatible with pouchdb-bulk-get project which was proposed as "standard implementation"

Ok, now I'm really confused now. If I make /_bulk_get compatible with PouchDB that will means that it will be not compatible with the Couchbase/RCouch and be very annoying for people who'll try use it outside of replication context (it's stupid to not).

If I start fixing oblivious flaws of proposed implementation, then we'll have yet another incompatible with the others /_bulk_get implementation. Not cool.

At this moment I want to flip the table and make this feature from scratch, starting first from the actual proposal about the API of final implementation, finding agreement on this in CouchDB team and call PouchDB and Couchbase folks to the table (once I put it back) for the further implementation as a part of replication API. And make this short and quickly since @dholth made all the work, current devil is in the details and shape of things.

Otherwise we (CouchDB) will end with semi-compatible with *ouchDB world resource that's broken for everyday usage. I'm -1 on such turn of events.

Certainly, I may be wrong at all of this, so any ideas are welcome.

Nudge list: @nolanlawson @dholth @janl @benoitc

dholth:
I may be the only pouchdb bulk get user. A top level array is a little
better for streaming. Does get return json for 404? That is what bulk
get should do.


On Tue, Apr 21, 2015, at 08:52 AM, Alexander Shorin wrote:
> Hi there, short update.


> I'd started porting this PR over 2.0. First note that in current
> state it doesn't works and breaks any document update, so don't try
> to merge it (:


> Second, while fixing it I found that initial format doesn't assumes
> any possible errors:


> $ echo '{"docs": [{"id": "foo"}, {"id": "bar"}]}' | http
> http://localhost:15986/test/_bulk_get
[ [{"_id": "foo", "_rev": "1-967a00dff5e02add41819138abb3284d"}],
[{"_id": "bar", "_rev": "1-6aaf7080aad9d1a9482e07c46581dac0"}] ]
>
> What the response should be if document id is missed or wrong or
> document not exists or some parameters are malformed? Oblivious
> solution is:


> $ echo '{"docs": [{"id": "foo"}, {"id": "bar"}]}' | http
> http://localhost:15986/test/_bulk_get
[ [{"_id": "foo", "_rev": "1-967a00dff5e02add41819138abb3284d"}],
[{"error": "not_found", "id": "bar", "reason": "not today"}] ]
>
> But such array of arrays makes me sad panda.


> I'd started to look on current implementations and found no agreement
> on what the response should looks like:


>  * Couchbase[1] assumes the response to be always multipart and
>    request payload should contain array of objects where "id" field is
>    mandatory, "rev" and "atts_since" are optional and no mention of
>    the others. So, passing open_revs will not work for Couchbase or
>    make it incompatible with.


>  * RCouch[2] returns JSON response in the fashion {"results": [{"id":
>    "foo", "docs": [...]}, ...]} what is completely different from
>    proposed patch and pouchdb-bulk-get implementation.


>  * Each of these are incompatible with pouchdb-bulk-get project which
>    was proposed as "standard implementation"


> Ok, now I'm really confused now. If I make /_bulk_get compatible with
> PouchDB that will means that it will be not compatible with the
> Couchbase/RCouch and be very annoying for people who'll try use it
> outside of replication context (it's stupid to not).


> If I start fixing oblivious flaws of proposed implementation, then
> we'll have yet another incompatible with the others /_bulk_get
> implementation. Not cool.


> At this moment I want to flip the table and make this feature from
> scratch, starting first from the actual proposal about the API of
> final implementation, finding agreement on this in CouchDB team and
> call PouchDB and Couchbase folks to the table (once I put it back) for
> the further implementation as a part of replication API. And make this
> short and quickly since @dholth[3] made all the work, current devil is
> in the details and shape of things.


> Otherwise we (CouchDB) will end with semi-compatible with *ouchDB
> world resource that's broken for everyday usage. I'm -1 on such turn
> of events.


> Certainly, I may be wrong at all of this, so any ideas are welcome.


> Nudge list: @nolanlawson[4] @dholth[5] @janl[6] @benoitc[7]


> — Reply to this email directly or view it on GitHub[8].

>



Links:

  1. http://developer.couchbase.com/mobile/develop/references/sync-gateway/rest-api/database/post-bulk-get/index.html
  2. https://github.com/rcouch/couchdb-couch-httpd/blob/master/src/couch_httpd_bulk_get.erl#L45
  3. https://github.com/dholth
  4. https://github.com/nolanlawson
  5. https://github.com/dholth
  6. https://github.com/janl
  7. https://github.com/benoitc
  8. https://github.com/apache/couchdb-couch/pull/18#issuecomment-94781101

dholth:
However assuming RCouch has users why not do their json. Double check to
see what arguments PouchDB _bulk_get actually uses.


On Tue, Apr 21, 2015, at 09:08 AM, Daniel Holth wrote:
> I may be the only pouchdb bulk get user. A top level array is a little
> better for streaming. Does get return json for 404? That is what bulk
> get should do.
>
>
> On Tue, Apr 21, 2015, at 08:52 AM, Alexander Shorin wrote:
>> Hi there, short update.


>> I'd started porting this PR over 2.0. First note that in current
>> state it doesn't works and breaks any document update, so don't try
>> to merge it (:


>> Second, while fixing it I found that initial format doesn't assumes
>> any possible errors:


>> $ echo '{"docs": [{"id": "foo"}, {"id": "bar"}]}' | http
>> http://localhost:15986/test/_bulk_get
[ [{"_id": "foo", "_rev": "1-967a00dff5e02add41819138abb3284d"}],
[{"_id": "bar", "_rev": "1-6aaf7080aad9d1a9482e07c46581dac0"}] ]
>>
>> What the response should be if document id is missed or wrong or
>> document not exists or some parameters are malformed? Oblivious
>> solution is:


>> $ echo '{"docs": [{"id": "foo"}, {"id": "bar"}]}' | http
>> http://localhost:15986/test/_bulk_get
[ [{"_id": "foo", "_rev": "1-967a00dff5e02add41819138abb3284d"}],
[{"error": "not_found", "id": "bar", "reason": "not today"}] ]
>>
>> But such array of arrays makes me sad panda.


>> I'd started to look on current implementations and found no agreement
>> on what the response should looks like:


>>  * Couchbase[1] assumes the response to be always multipart and
>>    request payload should contain array of objects where "id" field
>>    is mandatory, "rev" and "atts_since" are optional and no mention
>>    of the others. So, passing open_revs will not work for Couchbase
>>    or make it incompatible with.


>>  * RCouch[2] returns JSON response in the fashion {"results": [{"id":
>>    "foo", "docs": [...]}, ...]} what is completely different from
>>    proposed patch and pouchdb-bulk-get implementation.


>>  * Each of these are incompatible with pouchdb-bulk-get project which
>>    was proposed as "standard implementation"


>> Ok, now I'm really confused now. If I make /_bulk_get compatible with
>> PouchDB that will means that it will be not compatible with the
>> Couchbase/RCouch and be very annoying for people who'll try use it
>> outside of replication context (it's stupid to not).


>> If I start fixing oblivious flaws of proposed implementation, then
>> we'll have yet another incompatible with the others /_bulk_get
>> implementation. Not cool.


>> At this moment I want to flip the table and make this feature from
>> scratch, starting first from the actual proposal about the API of
>> final implementation, finding agreement on this in CouchDB team and
>> call PouchDB and Couchbase folks to the table (once I put it back)
>> for the further implementation as a part of replication API. And make
>> this short and quickly since @dholth[3] made all the work, current
>> devil is in the details and shape of things.


>> Otherwise we (CouchDB) will end with semi-compatible with *ouchDB
>> world resource that's broken for everyday usage. I'm -1 on such turn
>> of events.


>> Certainly, I may be wrong at all of this, so any ideas are welcome.


>> Nudge list: @nolanlawson[4] @dholth[5] @janl[6] @benoitc[7]


>> — Reply to this email directly or view it on GitHub[8].

>>
>



Links:

  1. http://developer.couchbase.com/mobile/develop/references/sync-gateway/rest-api/database/post-bulk-get/index.html
  2. https://github.com/rcouch/couchdb-couch-httpd/blob/master/src/couch_httpd_bulk_get.erl#L45
  3. https://github.com/dholth
  4. https://github.com/nolanlawson
  5. https://github.com/dholth
  6. https://github.com/janl
  7. https://github.com/benoitc
  8. https://github.com/apache/couchdb-couch/pull/18#issuecomment-94781101

kxepal:
> I may be the only pouchdb bulk get user. A top level array is a little better for streaming. 

Agree about streaming. 

> Does get return json for 404? That is what bulk get should do.

It cannot, because of streaming nature. Like /_bulk_docs, it sends response status 200 and headers before any payload processing (otherwise you'll loose streaming nature) what means that in case of any errors you cannot change response status to 404 / 500 or whatever. All you can is to send error as object in the response stream (as /_bulk_docs does). However, PouchDB-Get is not able to handle such case and will count these errors as docs.

kxepal:
> However assuming RCouch has users why not do their json. Double check to see what arguments PouchDB _bulk_get actually uses.

Well, I'm fine to break compatibility, but I wonder for which side I have to do this.

dholth:
The principle behind my _bulk_get is that it is *exactly* lots of GET
returned in an array, the first implementation was even written as a lua
intermediary between nginx and CouchDB.

This "exactly GET" property makes _bulk_get easier to document and
understand. All other implementations seemed to want to remove some of
the weird GET options that no one understands.

For a 404, the JSON response is:

{"error":"not_found","reason":"missing"}

So _bulk_get would return exactly that and maybe add an id field, or you
might have to remember the order of the original request.

[ [{"_id": "foo", "_rev": "1-967a00dff5e02add41819138abb3284d"}],
[{"error": "not_found", "_id": "bar", "reason": "missing"}] ]

I checked, and unfortunately my _bulk_get appears to fail entirely if an
element is not found.

I'm not sure about the array of arrays. Does GET sometimes return an
array, and sometimes it does not?

Thanks!

On Tue, Apr 21, 2015, at 09:22 AM, Alexander Shorin wrote:
>> However assuming RCouch has users why not do their json. Double check
>> to see what arguments PouchDB _bulk_get actually uses.


> Well, I'm fine to break compatibility, but I wonder for which side I
> have to do this.


> — Reply to this email directly or view it on GitHub[1].

>



Links:

  1. https://github.com/apache/couchdb-couch/pull/18#issuecomment-94792422

kxepal:
@dholth About the errors...how do you feel with https://github.com/couchbase/sync_gateway/wiki/Bulk-GET ?

Specifically:
```
If there's an error getting a document revision, most likely because it doesn't exist, its corresponding JSON body in the response will contain only the properties "id", "error", "reason" and "status", just as in a response from _all_docs.
```

Thought status isn't applicable for CouchDB nor PouchDB (afaik).

>  I'm not sure about the array of arrays. Does GET sometimes return an array, and sometimes it does not? 

Yes, if you request document with open_revs query param and explicit application/json Accept header.

nolanlawson:
> If I make /_bulk_get compatible with PouchDB

As @dholth said, his implementation is a *PouchDB plugin*. PouchDB does not depend on it.

> A top level array is a little better for streaming. 

The streaming stuff (pouchdb-replication-stream, pouchdb-load, socket-pouch, etc.) is a parallel effort . I consider it a long-term plan, whereas `_bulk_get`/`_bulk_revs` is the short-term fix.

> Well, I'm fine to break compatibility,

The stance of PouchDB (and I think @daleharvey would agree) is that we will implement whatever CouchDB implements. CouchDB is our gold standard, not RCouch or Couchbase.

That being said, if Couchbase has a solution that handles errors (even if a bit inelegantly), then I think we should prefer their version instead of inventing a third version. RCouch has few users (AFAIK), but Couchbase Mobile is pretty popular for iOS/Android development. So we would benefit by tapping in to their ecosystem.

PouchDB's only point of contention with the Couchbase API is that we need a JSON response rather than a multipart/mixed response. Basically if the client says `Accept: application/json`, then the server has to return JSON.

kxepal:
@nolanlawson Thanks for the input. Ok, seems things getting simpler now, since I was confused different implementations. Good, I'll stay with Couchbase implementation then with RCouch improvements to support application/json response. Basically, that means to backport RCouch implementation (: And everyone should be happy then.

nolanlawson:
:+1:

kxepal:
@dholth Hi! could you close your PR here please? This feature is available via chttpd.

dholth:
Hooray for improved replication.

PR #23, opened 2014-12-12
Add IS_OLD_RECORD macro

davisp:
This can be used in guards to detect if a state variable needs to be
upgraded. An example of it's use might be:

```
handle_call(Msg, From, St) when ?IS_OLD_RECORD(St, #st{}) ->
    handle_call(Msg, From, upgrade_state(St));
```

This is useful for upgrading any of our various gen behaviors that
aren't properly supervised.

COUCHDB-2511


kxepal:
@davisp ok with merge?

davisp:
+1 to merge. I guess I should probably get to that.

PR #24, opened 2014-12-15
2516 deduplicate attachements on compaction

iilyak:
This is replacing PR https://github.com/apache/couchdb-couch/pull/22


iilyak:
There are problems with calculating active size when document get updated or deleted.

davisp:
For posterity, we were trying to make this work without requiring an on-disk format change but sadly that appears to not be possible due to the lack of information available when updating documents. Given the lack of information we can't tell if an attachment that's being updated or deleted needs to be counted in active_size.

The proper fix for this will be to add a new index to each database that contains the attachment data. Given the merge work we're going to delay this until later. Also if/when we get back to this we'll also want to investigate having attachment data stored in a separate file which would remove a lot of the gymnastics we have around couch_file:make_blocks and friends.

PR #27, opened 2014-12-19
Modify init_delete_dir to be async

iilyak:
In the case when .delete directory contains lots of files.
The deletion of the directory leads to timeout in the supervisor.
Which in it's turn kills the node.
So we rename original directory so couchdb cannot find it.
Then we remove files from renamed directory asynchronously.

COUCHDB-2524

### DIFF
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -260,10 +260,13 @@ init_delete_dir(RootDir) ->
     % note: ensure_dir requires an actual filename companent, which is the
     % reason for "foo".
     filelib:ensure_dir(filename:join(Dir,"foo")),
-    filelib:fold_files(Dir, ".*", true,
-        fun(Filename, _) ->
-            ok = file:delete(Filename)
-        end, ok).
+    spawn(fun() ->
+        filelib:fold_files(Dir, ".*", true,
+            fun(Filename, _) ->
+                ok = file:delete(Filename)
+            end, ok)
+    end),
+    ok.
 
 
 read_header(Fd) ->


rnewson:
I'm confused. Why rename the directory if you're going to make the delete async? It seems the entire fix is just to spawn a process to do the cleanup of the .delete directory rather than block.

src/couch_file.erl
```diff
@@ -257,13 +257,43 @@ nuke_dir(RootDelDir, Dir) ->
...
+
+
+delete_temp_dirs(RootDir, Prefix) ->
+    Dirs = filelib:wildcard(Prefix ++ ".tmp-" ++ "*", RootDir),
+    [ok = delete_dir(Dir) || Dir <- Dirs],
+    ok.
+
+
+ensure_dir(Dir) ->
+    filelib:ensure_dir(Dir ++ "/").
```
rnewson: it's not always a forward slash ;)

kxepal: Yes, but windows works correctly with both of them actually.

iilyak: I though couchdb uses the files from .delete directory for recovery from crash. It looks like I was wrong.

iilyak: According to docs for filename module:
In Windows, all functions return file names with forward slashes only, even if the arguments contain back slashes. 
Also if I remember correctly it worked on windows. However it was long time ago when I did erlang project on windows. So it might be broken on recent erlang versions, but I doubt.

rnewson: anything in .delete can be deleted. Prior to the .delete dir, couchdb just called 'rm'. .delete exists so we can delete asynchronously rather than block. It's definitely not used for recovery.

Given this, I think the PR is far simpler? Just spawn the process that deletes the contents of the .delete file.

kxepal:
Ha! Plain and simple (: +1
@iilyak could you squash all your commits?

iilyak:
History is rewritten for merge

kxepal:
\o/

PR #29, opened 2015-01-06
2534 db security respect authed users

kxepal:
This does two things:
- Fixes the way how user name is searched in list. lists:member/2 made in BIF, faster and doesn't causes redundant list copy;
- Fixes readers vs members terminology in code;
- Fixes COUCHDB-2534, throwing forbidden error if user is authenticated, but is not a database member.


kxepal:
On second thought, using ordsets might be better choice for all these names/roles checks.

kxepal:
Applied refactoring onto this PR. Also throw correct error for check_is_admin/1 when end user is authed.

davisp:
+1 after squashing the refactor commit

PR #31, opened 2015-01-14
Add a configurable timeout for get_proc calls

mikewallace1979:
Previously the gen_server calls to couch_proc_manager/get_proc
used a timeout of infinity. There are multiple places in the
couch_proc_manager code path where that process can die without
replying. With an infinity timeout the couch_query_server process
would then hang around forever.

This commit makes the gen_server call to get_proc use a configurable
timeout, set by the couchdb/get_proc_timeout config variable. It
defaults to 60000ms.

Closes:

  COUCHDB-2425
  COUCHDB-2426

### DIFF
--- a/src/couch_query_servers.erl
+++ b/src/couch_query_servers.erl
@@ -342,9 +342,12 @@ proc_set_timeout(Proc, Timeout) ->
     {Mod, Func} = Proc#proc.set_timeout_fun,
     apply(Mod, Func, [Proc#proc.pid, Timeout]).
 
+get_os_process_timeout() ->
+    list_to_integer(config:get("couchdb", "os_process_timeout", "5000")).
+
 get_ddoc_process(#doc{} = DDoc, DDocKey) ->
     % remove this case statement
-    case gen_server:call(couch_proc_manager, {get_proc, DDoc, DDocKey}, infinity) of
+    case gen_server:call(couch_proc_manager, {get_proc, DDoc, DDocKey}, get_os_process_timeout()) of
     {ok, Proc, {QueryConfig}} ->
         % process knows the ddoc
         case (catch proc_prompt(Proc, [<<"reset">>, {QueryConfig}])) of
@@ -360,7 +363,7 @@ get_ddoc_process(#doc{} = DDoc, DDocKey) ->
     end.
 
 get_os_process(Lang) ->
-    case gen_server:call(couch_proc_manager, {get_proc, Lang}, infinity) of
+    case gen_server:call(couch_proc_manager, {get_proc, Lang}, get_os_process_timeout()) of
     {ok, Proc, {QueryConfig}} ->
         case (catch proc_prompt(Proc, [<<"reset">>, {QueryConfig}])) of
         true ->


kxepal:
`get_proc_timeout` is good name for internals, but for configuration it should be more user friendly and explain what exactly it controls. Exposing knowledge of gen_server calls to user level configs doesn't simplifies all the things (:

kxepal:
How about to reuse `couch_util:get_value(<<"timeout">>, QueryConfig)` for `get_proc` call?

mikewallace1979:
> get_proc_timeout is good name for internals, but for configuration it should be more user friendly and explain what exactly it controls. Exposing knowledge of gen_server calls to user level configs doesn't simplifies all the things

I agree but I couldn't figure out a better name given that it is pretty much only an internal gen_server concern. Will ponder some more. Maybe it's a sign that exposing it externally is a Bad Idea.

mikewallace1979:
> How about to reuse couch_util:get_value(<<"timeout">>, QueryConfig) for get_proc call?

I'd sort of assumed that we wouldn't want to use the same timeout which the QueryConfig picks up (which will be `couchdb/os_process_timeout`) since this was previously an infinity timeout. Also we don't get the QueryConfig record proplist until the get_proc has received a response.

mikewallace1979:
@kxepal Maybe re-using couchdb/os_process_timeout is actually the best way forward - solves the naming problem and means we're using one timeout for all os process related things, which is simpler from a config point of view. What do you think?

kxepal:
@mikewallace1979 I think it's good idea!

mikewallace1979:
Done.

Since the resulting timeout will be thrown back up to chttpd I think we'll also want https://github.com/apache/couchdb-chttpd/pull/18 so that instead of returning:

```
{"error":"timeout","reason":"{gen_server,call,[couch_proc_manager,{get_proc,<<\"javascript\">>},5000]}"}
```

to the client we instead return our normal timeout message of:

```
{"error":"timeout","reason":"The request could not be processed in a reasonable amount of time."}
```

kxepal:
@mikewallace1979 nice idea, but the message hides original issue: what exactly caused timeout error. May be just reformat reason in more human-friendly way?

mikewallace1979:
@kxepal The cause of the timeout error will be logged internally by rexi_server so the information is available to the operators. I was reasoning that as the client is external it can't really do anything with the details of what went wrong internally though I guess there's no reason why we shouldn't provide that information.

kxepal:
@mikewallace1979 Oh, in this case cool and agreed. All good for me, thanks a lot! Happy to merge, but let's wait for a while for others opinion.

kxepal:
@mikewallace1979 could you please squash your commits so I merge them?

janl:
just a note that “the operators“, at least in the beginning, are end-users as well, so we want to make this as friendly as possible.

mikewallace1979:
@kxepal All squashed - sorry for taking my time about it.

@janl +1 on friendliness, though I admit I find it's sometimes a hard thing to achieve in practice :) In this case I think returning the same message we return for all other timeouts is sufficiently friendly - in which case I'll re-open the chttpd [PR](https://github.com/apache/couchdb-chttpd/pull/18).

kxepal:
Landed, thanks! (:

PR #32, opened 2015-01-19
Make sure we load all applications with stats defs

iilyak:
couch_stats app relies on application:loaded_applications.
While in the test suite we have to load the apps manually
prior to start couch_stats app

COUCHDB-2552

### DIFF
--- a/src/test_util.erl
+++ b/src/test_util.erl
@@ -20,8 +20,7 @@
 -export([request/3, request/4]).
 -export([start_couch/0, start_couch/1, start_couch/2, stop_couch/0, stop_couch/1]).
 -export([start_config/1, stop_config/1]).
--export([start_applications/1]).
-
+-export([start_applications/1, stop_applications/1]).
 
 srcdir() ->
     code:priv_dir(couch) ++ "/../../".
@@ -86,27 +85,22 @@ start_couch(ExtraApps) ->
     start_couch(?CONFIG_CHAIN, ExtraApps).
 
 start_couch(IniFiles, ExtraApps) ->
+    load_applications_with_stats(),
     ok = application:set_env(config, ini_files, IniFiles),
     ok = lager:start(),
     ok = start_applications([inets, ibrowse, ssl, config, couch] ++ ExtraApps),
-    couch_stats:reload(),
     ok.
 
+stop_applications(Apps) ->
+    [ok = application:stop(App) || App <- lists:reverse(Apps)],
+    ok.
 
 stop_couch() ->
-    ok = application:stop(couch),
-    ok = application:stop(lager),
-    ok = application:stop(goldrush),
-    ok = application:stop(config),
-    ok = application:stop(ssl),
-    ok = application:stop(ibrowse),
-    ok = application:stop(inets),
+    ok = stop_applications([inets, ibrowse, ssl, config, goldrush, lager, couch]),
     ok.
-
 
 stop_couch(_) ->
     stop_couch().
-
 
 start_applications([]) ->
     ok;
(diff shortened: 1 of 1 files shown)


kxepal:
Yes, this fixes the issue. However, not sure about the fix correctness.

`start_couch` is basically starts minimal CouchDB instance with the specified config files (if any). Additionally, it may be instructed to start some more apps which implements some features, like: couch_mrview, couch_replicator etc.

This fix tries to break this and actually start minimal CouchDB instance + all the apps with the stats definition. This isn't minimalistic instance anymore, but not yet complete one.

So the question is: do we ever need in such startup definition (minimal core+extra apps)? May be we need to start CouchDB with all the apps all the time? Or even be more flexible: provide `start_minimal_couch([IniFiles, [ExtraApps]])` and `start_full_couch([IniFiles])` - what does it gives us? /cc @chewbranca 

The more finest fix for the 2552 is to touch couch_mrview tests setup function to call `start_couch` with `[couch_mrview]` extra app like couch_replicator does.

iilyak:
@kxepal [application:load/1 <> application:start/1](https://github.com/iilyak/couchdb-couch/blob/2552-unknown-metric-mrview-map_docs/src/test_util.erl#L144)

iilyak:
While working on test suite I had a need for start_couch(ExtraApps) and stop_couch(ExtraApps). I figure out that this API actually not side-effect free. Since when we do start_couch(Apps) we start dependencies. I tend to change it to start_couch/1 -> [started_apps]. So we can pass list of apps to teardown function. I think it is a sensible change. However this will affect lot's of repositories. There is an alternative to it which is to start a named process which would act as a term storage for the current test. Store arbitrary terms in the state of that process including info about apps that needs to be stopped on teardown. This would reduce number of arguments we would need to pass to tests themselves. We also can use unnamed process and pass Pid of it to every test as a Context and have test_util:get(Context, Key) and [Arg1, Arg2] = test_util:get(Context, [key1, key2]) to extract args from context. Not sure about the best approach to this. The change is out of the scope of this ticket anyway.

iilyak:
@kxepal BTW I've tried to use start_couch with [couch_mrview] extra app. It doesn't work in this case due to complex dependencies between `couch_stats`, `couch_logger`, `couch_index` and `couch_mrview`.

kxepal:
>  BTW I've tried to use start_couch with [couch_mrview] extra app. It doesn't work in this case due to complex dependencies between couch_stats, couch_logger, couch_index and couch_mrview.

Works for me. Example change:

```diff
diff --git a/src/couch_mrview_updater.erl b/src/couch_mrview_updater.erl
index db5b5e9..81b0085 100644
--- a/src/couch_mrview_updater.erl
+++ b/src/couch_mrview_updater.erl
@@ -187,7 +187,7 @@ map_docs(Parent, State0) ->
                 ({Id, Seq, Rev, deleted}, {SeqAcc, Results}) ->
                     {erlang:max(Seq, SeqAcc), [{Id, Seq, Rev, []} | Results]};
                 ({Id, Seq, Rev, Doc}, {SeqAcc, Results}) ->
-                    couch_stats:increment_counter([couchdb, mrview, map_docs],
+                    couch_stats:increment_counter([couchdb, mrview, map_doc],
                                                   1),
                     {ok, Res} = couch_query_servers:map_doc_raw(QServer, Doc),
                     {erlang:max(Seq, SeqAcc), [{Id, Seq, Rev, Res} | Results]}
diff --git a/test/couch_mrview_red_views_tests.erl b/test/couch_mrview_red_views_tests.erl
index 8c2ff47..2a07bcc 100644
--- a/test/couch_mrview_red_views_tests.erl
+++ b/test/couch_mrview_red_views_tests.erl
@@ -18,6 +18,9 @@
 -define(TIMEOUT, 1000).
 
 
+start()->
+    test_util:start_couch([couch_mrview]).
+
 setup() ->
     {ok, Db} = couch_mrview_test_util:init_db(?tempdb(), red),
     Db.
@@ -33,7 +36,7 @@ reduce_views_test_() ->
         "Reduce views",
         {
             setup,
-            fun test_util:start_couch/0, fun test_util:stop_couch/1,
+            fun start/0, fun test_util:stop_couch/1,
             {
                 foreach,
                 fun setup/0, fun teardown/1,
```

iilyak:
stale PR

PR #36, opened 2015-02-19
Consume JSON body even when its content is ignored

samueltardieu:
"_ensure_full_commit" and "_compact" require that the content-type
be set to "application/json". However, when some content is provided
(such as "{}" in order to send a valid JSON external representation
for the empty object), CouchDB needs to ensure that this content
is read from the incoming HTTP request.

Mirrors the fix for COUCHDB-2583 in CouchDB 1.x.


kxepal:
Thanks! (:

PR #37, opened 2015-02-24
Add couch_util:validate_callback_exists

iilyak:
COUCHDB-2585

PR #38, opened 2015-02-25
2625 dbname validator

iilyak:
This depends on https://github.com/apache/couchdb-couch/pull/37

### DIFF
--- a/src/couch_httpd.erl
+++ b/src/couch_httpd.erl
@@ -837,7 +837,7 @@ error_info({bad_ctype, Reason}) ->
     {415, <<"bad_content_type">>, Reason};
 error_info(requested_range_not_satisfiable) ->
     {416, <<"requested_range_not_satisfiable">>, <<"Requested range not satisfiable">>};
-error_info({error, illegal_database_name, Name}) ->
+error_info({error, {illegal_database_name, Name}}) ->
     Message = "Name: '" ++ Name ++ "'. Only lowercase characters (a-z), "
         ++ "digits (0-9), and any of the characters _, $, (, ), +, -, and / "
         ++ "are allowed. Must begin with a letter.",
--- a/src/couch_server.erl
+++ b/src/couch_server.erl
@@ -21,6 +21,7 @@
 -export([handle_cast/2,code_change/3,handle_info/2,terminate/2]).
 -export([dev_start/0,is_admin/2,has_admins/0,get_stats/0]).
 -export([close_lru/0]).
+-export([validate_dbname/1]).
 
 % config_listener api
 -export([handle_config_change/5, handle_config_terminate/3]).
@@ -36,7 +37,8 @@
     dbs_open=0,
     start_time="",
     update_lru_on_read=true,
-    lru = couch_lru:new()
+    lru = couch_lru:new(),
+    dbname_validator
     }).
 
 dev_start() ->
@@ -116,6 +118,9 @@ create(DbName, Options0) ->
 delete(DbName, Options) ->
     gen_server:call(couch_server, {delete, DbName, Options}, infinity).
 
+validate_dbname(DbName) ->
+    gen_server:call(couch_server, {validate_dbname, DbName}).
+
 maybe_add_sys_db_callbacks(DbName, Options) when is_binary(DbName) ->
     maybe_add_sys_db_callbacks(?b2l(DbName), Options);
 maybe_add_sys_db_callbacks(DbName, Options) ->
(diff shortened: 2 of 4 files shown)


src/couch_httpd.erl
```diff
@@ -837,7 +837,7 @@ error_info({bad_ctype, Reason}) ->
     {415, <<"bad_content_type">>, Reason};
 error_info(requested_range_not_satisfiable) ->
     {416, <<"requested_range_not_satisfiable">>, <<"Requested range not satisfiable">>};
-error_info({error, illegal_database_name, Name}) ->
+error_info({error, {illegal_database_name, Name}}) ->
     Message = "Name: '" ++ Name ++ "'. Only lowercase characters (a-z), "
```
kxepal: In case of db name validation customization this error message shouldn't be defined here since the requirements for the name could be different.

src/couch_server.erl
```diff
@@ -145,6 +150,8 @@ maybe_add_sys_db_callbacks(DbName, Options) ->
 path_ends_with(Path, Suffix) ->
     Suffix == lists:last(binary:split(mem3:dbname(Path), <<"/">>, [global])).
 
+check_dbname(#server{dbname_validator={M, F, A}}, DbName) ->
+    M:F(DbName, A);
 check_dbname(#server{dbname_regexp=RegExp}, DbName) ->
```
kxepal: Why do we have to maintain two different ways to apply dbname validation logic (MFA vs regex)? May be drop dbname_regexp and replace it with the default MFA which points to the original check_dbname? Would be a bit more consistent.

src/couch_httpd.erl
iilyak: We could change error to be `{error, {illegal_database_name, Name, ErrorMessage}}`

kxepal: In case if this wouldn't break anything, ok.

kxepal:
Fixing code duplicity with fabric is good, but do we actually need in check_dbname customization? 

Say, I want to have Cyrillic database names and I put my validation function into the config file. Will the other couch parts be ready to work with such names? How chttpd/httpd will map urlencoded utf-8 database name with a file name if filesystem uses different charset, like utf-16 on windows? I have a strong fear that this will cause a server failure. And in this case why do I have to change the default database name validation?

janl:
I’m with @kxepal: sceptical about the usefulness of letting the end user define how to validate database names.

I’m not sure what problem this solves. The referenced issue in #37 (https://issues.apache.org/jira/browse/COUCHDB-2585) is about something else.

iilyak:
Would you accept a PR with configurable whitelist of DBs which would be consulted in couch_server:checkdb_name?

janl:
I’d like to understand the underlying issue this is trying to solve, first :)

janl:
I found the related ticket: https://issues.apache.org/jira/browse/COUCHDB-2625 — It doesn’t answer my question though

iilyak:
I need to extend this list https://github.com/apache/couchdb-couch/blob/master/src/couch_server.erl#L152:L153 without modifying couchdb source code

janl:
@iilyak for what reason? :) (sorry for being dense)

kxepal:
hm... @iilyak I have a PR in stash that extends this list too by three more dbs (_nodes, _dbs, _metadata). But I ended with SYSTEM_DBS macro for now. 

However, the most simple way to maintain this list is using ini file. On other hand, these databases are special and not sure if it's ok when user have a power to add more than we actually need.

iilyak:
The reason to have it configurable is to allow vendor specific customization.

janl:
@iilyak why would you need a different set of db names? what are you trying to do there?

iilyak:
We have an existent dbs in the underscore namespace in our fork.

kxepal:
@iilyak so may be instead we make this list of system database names configurable while not expose ability to override default database name constraints? Sound as more solid approach.

iilyak:
@kxepal Following could solve the problem [based on](https://github.com/apache/couchdb-couch/pull/39/files)
```erlang
Whitelist = config:get("couchdb", "extra_system_dbs", []),
case lists:member(DbName, ?SYSTEM_DATABASES ++ Whitelist) of
```

kxepal:
mmm...yes, that certainly would works, but in long term personally I think about some some of dynamic registration of system databases: when couch_replicator loads, it registers _replicator database. When cassim app loads, it registers _metadata one and so on. Why? Because system databases plays some special role for CouchDB and users shouldn't register here their own to prevent possible collisions while these system databases are managed by some application which applies certain logic upon the data they holds. 

Speaking about your case, the solution would be trivial: make a plugin which registers more system databases and here you go.

kxepal:
How do you feel about such solution?

iilyak:
Yes this would work but it makes your PR and my PR obsolete. Since it is a late night for you @kxepal I can implement it. Closing this PR for that reason.

PR #42, opened 2015-03-29
allow downloading docs/views

robertkowalski:
allow downloading docs/views

enables downloading of views using a browser, e.g. via a button
in Fauxton.

api: http://localhost:5984/mydb/_design/foobar?download=true

PRs:
https://github.com/apache/couchdb-couch/pull/42
https://github.com/apache/couchdb/pull/310
https://github.com/apache/couchdb-chttpd/pull/30

COUCHDB-2433

### DIFF
--- a/src/couch_httpd_db.erl
+++ b/src/couch_httpd_db.erl
@@ -17,7 +17,8 @@
     db_req/2, couch_doc_open/4, handle_db_changes_req/2,
     update_doc_result_to_json/1, update_doc_result_to_json/2,
     handle_design_info_req/3, parse_copy_destination_header/1,
-    parse_changes_query/2, handle_changes_req/4]).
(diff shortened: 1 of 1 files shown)


kxepal:
In additional to "attachment" content disposition type, the filename attribute is strongly recommended if you want to make this feature really user frieldly. Otherwise, browsers will ask to download very strange things. For instance, you request `http://localhost:5984/db/doc` => browser will ask you to download file `doc`, while `doc.json` might be more friendly.

For views there is the same story, but I think there need to include ddoc name into their file as well: `http://localhost:5984/db/_design/orders/_view/by_date?limit=10&download=true` => browser asks to download file `orders-by_date.json`.

Dealing with `filename` attribute will require a bit fines since it's not friendly for unicode names. See http://greenbytes.de/tech/tc2231/ for how they are processed.

PR #49, opened 2015-05-14
Include lager into list of extra apps when testing

iilyak:
Starting lager as `lager:start` is an issue since dependencies
started by lager are not included in a list of started applications
so we don't stop them after we are done with current test.

This commit uses test_util:start_applications so all deps of lager are
included into #test_context{}

COUCHDB-2688


rnewson:
+1

PR #51, opened 2015-05-21
Add test suite for Etag generation in mrview

iilyak:
COUCHDB-2650


davisp:
You should adding and removing documents in between requests so that you can verify that the etag changes appropriately.

iilyak:
The work for fixes tested by this test suite need to be re-done

PR #52, opened 2015-05-23
Fixing a build for Mac OS 10.10

arturmazurek:
Missing libmozjs185.

Added /usr/local/lib to linker search path because
couchdb suggests using homebrew to install dependancies and that's
where they are put by default.

COUCHDB-2699

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -81,7 +81,7 @@ CompareSrc = ["priv/couch_ejson_compare/*.c"],
 
 BaseSpecs = [
         %% couchjs
-        {"darwin", CouchJSPath, CouchJSSrc, [{env, [{"CFLAGS", JS_CFLAGS ++ "-DXP_UNIX -I/usr/local/include/js"}, {"LDFLAGS", JS_LDFLAGS}]}]},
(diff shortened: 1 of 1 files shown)


kxepal:
Thanks for the patch! LGFM. @janl @robertkowalski could you confirm?

janl:
I don’t see this breaking without this patch.

arturmazurek:
Odd, definitely happens for me. Yosemite 10.10.3 and CC "Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)"

Calling `cc -Xlinker -v` prints only the following library search path:
`/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/usr/lib`

robertkowalski:
I also can not confirm that error - @arturmazurek did you install brew just for couchdb? do brew installs run fine for you in general? did you source your bash_profile after installing brew?

arturmazurek:
I've had brew for some time already and it works just fine. I didn't source my bash_profile, however have rebooted multiple times since installing, so I guess would be sourced anyway.

So, since it doesn't happen for you Guys, I'll just close it off, no point in wasting time on it. I know how to fix it for my build and if I research more and have something, I'll post another pull request. Thanks.

janl:
There is no harm in explicitly adding the search path, I’ll get this committed in a bit!

arturmazurek:
Oh, that's great, thanks.

janl:
merged, thank you! :)

PR #53, opened 2015-05-25
Fix duplicated Content-Type for show/update functions

arturmazurek:
When a show/update function returned data it added a
"Content-Type: application/json" header by itself. Then
couch_httpd added some default headers (including a yet
another "application/json" one).
This diff deduplicates those headers using a set.

COUCHDB-1876


kxepal:
Thanks for the patch! Can you provide tests for your changes? 

Also, for CouchDB 2.0 you fixed problem only on local interface, while for [cluster|https://github.com/apache/couchdb-chttpd/blob/master/src/chttpd.erl] one it still may exists. Worth to check.

arturmazurek:
Ok, I'll add the tests and also check the cluster interface.

PR #54, opened 2015-05-26
Fix `active_size` format conversion in `get_db_info` function

eiri:
In `active_size` conversion in `couch_db:get_db_info/1` old db reduction's size format assumed to be an integer representing active state, when it also could be a tuple of active and external sizes.

This handeled correctly, for example, in [couch_db_updater.erl](https://github.com/apache/couchdb-couch/blob/master/src/couch_db_updater.erl#L426).

This patch addresses conversion of both possible formats to CouchDB `size_info` record.

This closes issue COUCHDB-2701

### DIFF
--- a/src/couch_db.erl
+++ b/src/couch_db.erl
@@ -338,8 +338,8 @@ get_db_info(Db) ->
             SI;
         {AS, ES} ->
             #size_info{active=AS, external=ES};
-        SI ->
-            #size_info{active=SI}
+        AS ->
+            #size_info{active=AS}
     end,
     ActiveSize = active_size(Db, SizeInfo),
     DiskVersion = couch_db_header:disk_version(Header),
--- a/src/couch_db.erl
+++ b/src/couch_db.erl
@@ -333,8 +333,13 @@ get_db_info(Db) ->
     {ok, FileSize} = couch_file:bytes(Fd),
     {ok, DbReduction} = couch_btree:full_reduce(IdBtree),
     SizeInfo0 = element(3, DbReduction),
-    SizeInfo = if is_record(SizeInfo0, size_info) -> SizeInfo0; true ->
-        #size_info{active=SizeInfo0}
+    SizeInfo = case SizeInfo0 of
+        SI when is_record(SI, size_info) ->
+            SI;
+        {AS, ES} ->
+            #size_info{active=AS, external=ES};
+        SI ->
+            #size_info{active=SI}
     end,
     ActiveSize = active_size(Db, SizeInfo),
     DiskVersion = couch_db_header:disk_version(Header),
--- a/src/couch_db_updater.erl
+++ b/src/couch_db_updater.erl
@@ -516,13 +516,13 @@ reduce_sizes(nil, _) ->
     nil;
 reduce_sizes(_, nil) ->
     nil;
-reduce_sizes(S1, S2) when is_integer(S1); is_integer(S2) ->
-    reduce_sizes(upgrade_sizes(S1), upgrade_sizes(S2));
 reduce_sizes(#size_info{}=S1, #size_info{}=S2) ->
     #size_info{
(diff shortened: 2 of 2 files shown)


kxepal:
How it's possible to reproduce such case? Never saw any problem with sizes for db info.

eiri:
@kxepal I can reproduce it by migrating a test database from DBCore to CouchDB 2.0, but I doubt it could be shown in migration from CouchDB 1.6.x to 2.0, as far as I can tell from a code, 1.6.x doesn't have a concept of external size and just uses integer in there. I can try though.

kxepal:
@eiri by migrate you mean simple copy-paste db file from 1.x to 2.0 database directory? And what's the different between DBCore and CouchDB 1.x?

eiri:
@kxepal  I don't think simple copy would do, given the database file going to be named as a shard. I thought of something like creating test db for 1.x, creating empty db with same name in 2.x (probably as r,w,n,q = 1), then overwriting it with 1.x and running compaction. Tricky business, but should be doable.

DBCore and CouchDB 1.x are quite different, but DBCore and Couch 2.0 are essentially the same apart from active/external format. Couch 2.0 uses #active_size record for it and DBCore just a 2 elements tuple, hence the PR.

hdiedrich:
I was asking for where reduce = is populated for checking how the tuple gets into that place.

eiri:
@hdiedrich : in [couch_db_updater.erl](https://github.com/apache/couchdb-couch/blob/master/src/couch_db_updater.erl#L492)

It's the same in DBCore's `couch_db_updater:btree_by_id_reduce/2`, except, but of course, for the record vs tuple discrepancy.

hdiedrich:
I mean the `#btree`'s `reduce` field that is used in `couch_btree.erl`
```Erlang
full_reduce(#btree{root=nil,reduce=Reduce}) ->
   {ok, Reduce(reduce, [])};
full_reduce(#btree{root=Root}) ->
   {ok, element(2, Root)}.
```
and ultimately produces the tuple `{AS, ES}` at the heart of this PR, which could be in `SizeInfo0`, via 

```Erlang
{ok, DbReduction} = couch_btree:full_reduce(IdBtree),
SizeInfo0 = element(3, DbReduction),
```

eiri:
@hdiedrich Yes I understand the question and link in my previous answer is pointing to the code point. It's populated on `couch_db_updater` init and `btree_by_id_reduce` is the reducer.

hdiedrich:
I was wondering where it is being assigned to the `reduce` field, found it here, thanks! https://github.com/apache/couchdb-couch/blob/master/src/couch_db_updater.erl#L553, https://github.com/apache/couchdb-couch/blob/master/src/couch_btree.erl#L49

src/couch_db.erl
```diff
@@ -332,8 +332,13 @@ get_db_info(Db) ->
...
     {ok, DbReduction} = couch_btree:full_reduce(IdBtree),
     SizeInfo0 = element(3, DbReduction),
-    SizeInfo = if is_record(SizeInfo0, size_info) -> SizeInfo0; true ->
-        #size_info{active=SizeInfo0}
+    SizeInfo = case SizeInfo0 of
+        SI when is_record(SI, size_info) ->
+            SI;
+        {AS, ES} ->
+            #size_info{active=AS, external=ES};
+        SI ->
```
rnewson: this variable should be called AS for consistency.

rnewson:
+1 after variable name changed for consistency.

eiri:
@rnewson: I've made an another related update in `couch_db_updater.erl`. Can you please take a look?

@davisp: Can you please take a look at this PR if you have time? It does the changes that hard to verify without DBCore and I'd like to have another +1 before merging it.

davisp:
+1

rnewson:
+1

PR #55, opened 2015-05-26
Fix duplicated Content-Type for show/update functions

arturmazurek:
When a show/update function returned data it added a
"Content-Type: application/json" header by itself. Then
couch_httpd added some default headers (including a yet
another "application/json" one).
This diff will add default chttpd headers only when
same headers are not yet present in the response.

COUCHDB-1876

### DIFF
--- a/src/couch_httpd.erl
+++ b/src/couch_httpd.erl
@@ -719,23 +719,17 @@ send_json(Req, Code, Value) ->
 
 send_json(Req, Code, Headers, Value) ->
     initialize_jsonp(Req),
-    DefaultHeaders = [
-        {"Content-Type", negotiate_content_type(Req)},
-        {"Cache-Control", "must-revalidate"}
-    ],
+    AllHeaders = maybe_add_default_headers(Req, Headers),
     Body = [start_jsonp(), ?JSON_ENCODE(Value), end_jsonp(), $\n],
-    send_response(Req, Code, DefaultHeaders ++ Headers, Body).
+    send_response(Req, Code, AllHeaders, Body).
 
 start_json_response(Req, Code) ->
     start_json_response(Req, Code, []).
 
 start_json_response(Req, Code, Headers) ->
     initialize_jsonp(Req),
-    DefaultHeaders = [
-        {"Content-Type", negotiate_content_type(Req)},
-        {"Cache-Control", "must-revalidate"}
-    ],
-    {ok, Resp} = start_chunked_response(Req, Code, DefaultHeaders ++ Headers),
+    AllHeaders = maybe_add_default_headers(Req, Headers),
+    {ok, Resp} = start_chunked_response(Req, Code, AllHeaders),
     case start_jsonp() of
         [] -> ok;
         Start -> send_chunk(Resp, Start)
@@ -745,6 +739,19 @@ start_json_response(Req, Code, Headers) ->
 end_json_response(Resp) ->
     send_chunk(Resp, end_jsonp() ++ [$\n]),
     last_chunk(Resp).
+
+maybe_add_default_headers(ForRequest, ToHeaders) ->
+    DefaultHeaders = [
+        {"Content-Type", negotiate_content_type(ForRequest)},
+        {"Cache-Control", "must-revalidate"}
+    ],
+    lists:foldl(fun maybe_add_header/2, ToHeaders, DefaultHeaders).
+
+maybe_add_header({HeaderName, HeaderValue}, ToHeaders) ->
+    case lists:keyfind(HeaderName, 1, ToHeaders) of
+        false -> ToHeaders ++ [{HeaderName, HeaderValue}];
+        _Found -> ToHeaders
+    end.
 
 initialize_jsonp(Req) ->
     case get(jsonp) of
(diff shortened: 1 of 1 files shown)


arturmazurek:
This is a follow-up to https://github.com/apache/couchdb-couch/pull/53

Added tests for the new functions.

Also regarding a comment in that old pull request:
> Also, for CouchDB 2.0 you fixed problem only on local interface, while for [cluster|https://github.com/apache/couchdb-chttpd/blob/master/src/chttpd.erl] one it still may exists. Worth to check.

I was checking it through chttpd interface. - if I understand correctly `dev/run` starts a cluster listening at http://localhost:15984 and that interface reuses code from couch, for which this fix is.

src/couch_httpd.erl
```diff
@@ -745,6 +739,19 @@ start_json_response(Req, Code, Headers) ->
 end_json_response(Resp) ->
     send_chunk(Resp, end_jsonp() ++ [$\n]),
     last_chunk(Resp).
+	
+add_default_headers_if_needed(ForRequest, ToHeaders) ->
```
kxepal: `maybe_add_default_headers` sounds better, shorter and easy to read.

src/couch_httpd.erl
```diff
@@ -30,6 +30,7 @@
 -export([send_json/2,send_json/3,send_json/4,last_chunk/1,parse_multipart_request/3]).
 -export([accepted_encodings/1,handle_request_int/5,validate_referer/1,validate_ctype/2]).
 -export([http_1_0_keep_alive/2]).
+-export([add_default_headers_if_needed/2, add_header_if_needed/2]).
```
kxepal: If these functions are internal, no need to pollute export API. eunit tests could be defined inline at the end of the module via check against TEST definition.

test/couch_httpd_headers_tests.erl
```diff
@@ -0,0 +1,55 @@
...
+    % json serving method. Defaults to 'application/javascript' otherwise.
+    % In this test, with an empty dummy request it will also default to that.
+    DummyRequest = [],
+    NoCacheHeader = {"Cache-Control", "no-cache"},
+    DefaultMustRevalidateHeader = {"Cache-Control", "must-revalidate"}, 
+    DefaultApplicationJavascriptHeader = {"Content-Type", "application/javascript"},
+    ApplicationJsonHeader = {"Content-Type", "application/json"},
+    Cases = [
+        % When there are no headers should add Content-Type and Cache-Control
+        {[],                                     [DefaultApplicationJavascriptHeader, DefaultMustRevalidateHeader]},
```
kxepal: Can we not go beyond from 80 right margin?

test/couch_httpd_headers_tests.erl
```diff
@@ -0,0 +1,55 @@
...
+        % should add to non-empty header_list
+        {[{"K1", "V1"}],               {"K2", "V2"}, [{"K1", "V1"}, {"K2", "V2"}]},
+         % should not add or change when already there
+        {[{"K1", "V1"}],               {"K1", "V2"}, [{"K1", "V1"}]},
+         % should not add or change when already there
+        {[{"K1", "V1"}, {"K2", "V2"}], {"K1", "V2"}, [{"K1", "V1"}, {"K2", "V2"}]},
+        % should add to a bit longer list
+        {[{"K1", "V1"}, {"K2", "V2"}], {"K3", "V3"}, [{"K1", "V1"}, {"K2", "V2"}, {"K3", "V3"}]} 
+    ],
+    lists:map(fun({InitialHeaders, HeaderToAdd, ProperResult}) ->
```
kxepal: Here you need to use [tests generation](http://www.erlang.org/doc/apps/eunit/chapter.html#id61504) - scroll down to "Writing test generating functions".

test/couch_httpd_headers_tests.erl
```diff
@@ -0,0 +1,55 @@
...
+        % When there are no headers should add Content-Type and Cache-Control
+        {[],                                     [DefaultApplicationJavascriptHeader, DefaultMustRevalidateHeader]},
+        % When there's already a Cache-Control header, should add only Content-Type header
+        {[NoCacheHeader],                        [NoCacheHeader, DefaultApplicationJavascriptHeader]},
+        % When there's already a Content-Type header should add Cache-Control header
+        {[ApplicationJsonHeader],                [ApplicationJsonHeader, DefaultMustRevalidateHeader]},
+        % When both Cache-Control and Content-Type headers are present shouldn't add anything
+        {[NoCacheHeader, ApplicationJsonHeader], [NoCacheHeader, ApplicationJsonHeader]}
+    ],
+    lists:map(fun({InitialHeaders, ProperResult}) ->
```
kxepal: Same test generation should be applied here.

kxepal:
Thanks! Basically, looks good, but left a few commentaries about the implementation.

src/couch_httpd.erl
arturmazurek: I am unsure if I get this one, do you mean adding something like the following at the end of this module? 

``` erlang
-ifdef(TEST).
add_default_headers_test() -> something.
-endif.
```
If yes, sure, will do.

kxepal: right

arturmazurek:
Changed what @kxepal pointed out. 
Sorry for the second update to the pull request today, my editor does tabs instead of spaces by default and I missed that...

arturmazurek:
// rebased

kxepal:
@arturmazurek thanks! Let me check it, but LGFM.

kxepal:
@arturmazurek Please fix the following issues in amend commit:

```
55.patch:54: trailing whitespace.
    
55.patch:85: trailing whitespace.
           "Adding to empty headers"},      
55.patch:105: trailing whitespace.
           "Adding header to 2 elements headers list"} 
55.patch:108: trailing whitespace.
        {Desc, 
55.patch:109: trailing whitespace.
        ?_assertEqual(ProperResult, 
warning: squelched 4 whitespace errors
warning: 9 lines add whitespace errors.
```

arturmazurek:
Should be fine - `git diff HEAD~1 --check` now outputs nothing

kxepal:
+1

kxepal:
@arturmazurek sorry for delay! I rebased your patch and merged. Thanks for contribution!

PR #58, opened 2015-06-15
Introduce couch_db:normalize_dbname

iilyak:
Move duplicated logic into couch_db:normalize_dbname. We could use this
helper function anywhere we need to extract dbname from shard path.

COUCHDB-2715

We would need this PR for COUCHDB-2635

### DIFF
--- a/src/couch_db.erl
+++ b/src/couch_db.erl
@@ -33,6 +33,7 @@
 -export([load_validation_funs/1]).
 -export([check_md5/2, with_stream/3]).
 -export([monitored_by/1]).
+-export([normalize_dbname/1]).
 
 -include_lib("couch/include/couch_db.hrl").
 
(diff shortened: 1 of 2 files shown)


kxepal:
+1

src/couch_db.erl
```diff
@@ -1469,3 +1470,8 @@ select_gt(V1, _V2) -> V1.
 
 select_lt(V1, V2) when V1 > V2 -> V2;
 select_lt(V1, _V2) -> V1.
+
+normalize_dbname(<<"shards/", _/binary>> = Path) ->
+    lists:last(binary:split(mem3:dbname(Path), <<"/">>, [global]));
```
kxepal: Side question: isn't lists:reverse+lists:head better than lists:last?

rnewson: for very short lists like this, I doubt it'll matter.

PR #60, opened 2015-06-23
Configurable password scheme

rnewson:
This gives the administrator control over which algorithm is used to
hash passwords and a separate control over whether this happens on
successful authentication or only at password change time.

closes COUCHDB-2725

### DIFF
--- a/src/couch_httpd_auth.erl
+++ b/src/couch_httpd_auth.erl
@@ -372,9 +372,11 @@ maybe_value(Key, Else, Fun) ->
 
 maybe_upgrade_password_hash(Req, UserName, Password, UserProps,
         AuthModule, AuthCtx) ->
+    Upgrade = config:get_boolean("couch_httpd_auth", "upgrade_password_on_auth", true),
     IsAdmin = lists:member(<<"_admin">>, couch_util:get_value(<<"roles">>, UserProps, [])),
-    case {IsAdmin, couch_util:get_value(<<"password_scheme">>, UserProps, <<"simple">>)} of
-    {false, <<"simple">>} ->
+    case {IsAdmin, Upgrade,
+         couch_util:get_value(<<"password_scheme">>, UserProps, <<"simple">>)} of
+    {false, true, <<"simple">>} ->
         UserProps2 = proplists:delete(<<"password_sha">>, UserProps),
         UserProps3 = [{<<"password">>, Password} | UserProps2],
         NewUserDoc = couch_doc:from_json_obj({UserProps3}),
--- a/src/couch_passwords.erl
+++ b/src/couch_passwords.erl
@@ -30,6 +30,15 @@ simple(Password, Salt) when is_binary(Password), is_binary(Salt) ->
 hash_admin_password(ClearPassword) when is_list(ClearPassword) ->
     hash_admin_password(?l2b(ClearPassword));
 hash_admin_password(ClearPassword) when is_binary(ClearPassword) ->
+    %% Support both schemes to smooth migration from legacy scheme
+    Scheme = config:get("couch_httpd_auth", "password_scheme", "pbkdf2"),
+    hash_admin_password(Scheme, ClearPassword).
+
+hash_admin_password("simple", ClearPassword) -> % deprecated
+    Salt = couch_uuids:random(),
+    Hash = crypto:sha(<<ClearPassword/binary, Salt/binary>>),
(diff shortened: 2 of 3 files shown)


src/couch_httpd_auth.erl
```diff
@@ -372,9 +372,11 @@ maybe_value(Key, Else, Fun) ->
 
 maybe_upgrade_password_hash(Req, UserName, Password, UserProps,
         AuthModule, AuthCtx) ->
+    Upgrade = config:get("couch_httpd_auth", "upgrade_password_on_auth", "true"),
```
kxepal: [config:get_boolean/3](https://github.com/apache/couchdb-config/blob/master/src/config.erl#L102-L108) is better here.

src/couch_passwords.erl
```diff
@@ -30,6 +30,14 @@ simple(Password, Salt) when is_binary(Password), is_binary(Salt) ->
 hash_admin_password(ClearPassword) when is_list(ClearPassword) ->
     hash_admin_password(?l2b(ClearPassword));
 hash_admin_password(ClearPassword) when is_binary(ClearPassword) ->
+    Scheme = config:get("couch_httpd_auth", "password_scheme", "pbkdf2"),
+    hash_admin_password(Scheme, ClearPassword).
+
+hash_admin_password("simple", ClearPassword) ->
+    Salt = couch_uuids:random(),
+    Hash = crypto:sha(<<ClearPassword/binary, Salt/binary>>),
+    ?l2b("-hashed-" ++ couch_util:to_hex(Hash) ++ "," ++ ?b2l(Salt));
```
kxepal: May be `lists:flatten(["-hashed-", couch_util:to_hex(Hash), ",", ?b2l(Salt)])`? Does lesser copies.

src/couch_users_db.erl
```diff
@@ -71,7 +82,9 @@ save_doc(#doc{body={Body}} = Doc) ->
         Body2 = ?replace(Body1, ?DERIVED_KEY, DerivedKey),
         Body3 = ?replace(Body2, ?SALT, Salt),
         Body4 = proplists:delete(?PASSWORD, Body3),
-        Doc#doc{body={Body4}}
+        Doc#doc{body={Body4}};
+    {_ClearPassword, _Scheme} ->
+        throw({forbidden, "Server cannot hash passwords at this time."})
```
kxepal: May be explicitly return response that specified Scheme is not supported now?

src/couch_passwords.erl
```diff
@@ -30,6 +30,14 @@ simple(Password, Salt) when is_binary(Password), is_binary(Salt) ->
 hash_admin_password(ClearPassword) when is_list(ClearPassword) ->
     hash_admin_password(?l2b(ClearPassword));
 hash_admin_password(ClearPassword) when is_binary(ClearPassword) ->
+    Scheme = config:get("couch_httpd_auth", "password_scheme", "pbkdf2"),
+    hash_admin_password(Scheme, ClearPassword).
+
+hash_admin_password("simple", ClearPassword) ->
```
janl: Is this re-introducing the salt+sha1 hash from the pre-pbkdf2 days or am I reading this wrong?

src/couch_passwords.erl
rnewson: I was deliberately following the previous code and I doubt this is a performance critical piece.

src/couch_users_db.erl
rnewson: I didn't want to expose the invalid value the administrator had set by mistake.

src/couch_passwords.erl
rnewson: not by default. If you change password_scheme to simple, then the server will continue to use that algorithm. This is all about migration.

src/couch_users_db.erl
kxepal: May be log then that fact? Otherwise such mistakes will remain unnoticed.

src/couch_passwords.erl
janl: How did we handle this in the past?

rnewson: we didn't, that's the enhancement. Plaintext admin passwords used to be one-pass sha'ed, then from 1.2 or so they were pbkdf2'd. Already hashed passwords are left as they are.

janl:
LGTM +1

PR #61, opened 2015-06-24
Remove compression's optimization

eiri:
When a file compression set to snappy, couch is doing an additional
optimization step by also compressing the term with deflate,
comparing the sizes of the result binary and choosing the smaller one.
This leads to a situation when for snappy compresed database the
'winning' deflate compressed term got decompressed and compressed
back into deflate on each document's write.

This patch removes this compression's optimization.

[Basic test](http://nbviewer.ipython.org/gist/eiri/79d91a797af9c6a6ff6d)
demonstrate that the gained with it disk space is not significant
enough to justify empty CPU cycles.

This closes COUCHDB-2726

### DIFF
--- a/src/couch_compress.erl
+++ b/src/couch_compress.erl
@@ -43,6 +43,8 @@ compress(<<?SNAPPY_PREFIX, _/binary>> = Bin, snappy) ->
     Bin;
 compress(<<?SNAPPY_PREFIX, _/binary>> = Bin, Method) ->
     compress(decompress(Bin), Method);
+compress(<<?COMPRESSED_TERM_PREFIX, _/binary>> = Bin, {deflate, _Level}) ->
+    Bin;
 compress(<<?TERM_PREFIX, _/binary>> = Bin, Method) ->
     compress(decompress(Bin), Method);
 compress(Term, none) ->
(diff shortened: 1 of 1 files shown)


janl:
nice one!

kxepal:
I believe this difference will only grow with database size, right?

P.S. Nice IPyNotebook usage here!

eiri:
@kxepal Thanks, stumble upon it recently, still wrapping my head around it :smile_cat:

The size diff is going to grow, sure, but in somehow glacier-like pace. I've made a quick comparison on 10000 docs database. There something like 200K size diff on optimization for 30MB of _pure_ data growth and delta itself seems to be linear. I've updated the notebook's gist with pretty graphs for the reference.

kxepal:
It seems the most safe is in CPU cycles which should be notable on weak hosts. But disk space reduction is a good small bonus (:

+1

davisp:
Ahh, subtle. Took me a few minutes of reading to realize that when the optimization kicks in we store the deflate version. On a subsequent read/write we realize its not snappy so we do the whole snappy compression again but then realize the snappy version is slightly bigger.

+1 to merge.

eiri:
Merged as per https://github.com/apache/couchdb-couch/commit/3a26ea1ba09e50da3b97b64e6e1ebf75c9406202

Closing

PR #62, opened 2015-06-26
Make sure authentication_db for clustered is considered as users db

iilyak:


rnewson:
Can you explain this patch?

Remembering that the clustered _users database and the node local _users databases are separate (a request to :5984 must not use the node-local _users, a request to :5986 must not use the clustered _users).

iilyak:
The explanation is in a bug description https://issues.apache.org/jira/browse/COUCHDB-2731

iilyak:
Let's say we have following in default.ini
```
[chttpd_auth] 
authentication_db = myusers
```
This would mean that it will not match on [any condition](https://github.com/apache/couchdb-couch/blob/master/src/couch_server.erl#L123:L124) we check to detect users db.
This would mean that `couch_users_db:before_doc_update` wouldn't be called. I.e. plain password upgrade will not happen. Since some other parts of couch expect valid profile which includes for example  `<<"password_scheme">>` it causes failures.

rnewson:
ok, I understand now. The fix here isn't quite right, I think as it will affect different databases as a side effect. If, say, the chttpd_auth db is called foo and the couch_http_db is called bar, then a node-local database called foo is also treated as a users db, as is a clustered database called bar.

iilyak:
@rnewson: PR is updated to address the problem you've identified.

rnewson:
re-reading this, I think it's a mistake for couch to know about chttpd config settings.

src/couch_server.erl
```diff
@@ -116,12 +116,16 @@ delete(DbName, Options) ->
...
+    Normalized = couch_db:normalize_dbname(DbName),
+    IsShard = couch_db:is_shard(DbName),
+
     DbsDbName = config:get("mem3", "shards_db", "_dbs"),
     NodesDbName = config:get("mem3", "nodes_db", "_nodes"),
-    IsReplicatorDb = DbName == config:get("replicator", "db", "_replicator") orelse
-	path_ends_with(DbName, <<"_replicator">>),
-    IsUsersDb = DbName ==config:get("couch_httpd_auth", "authentication_db", "_users") orelse
-	path_ends_with(DbName, <<"_users">>),
+    IsReplicatorDb = Normalized == config:get("replicator", "db", "_replicator") orelse
```
davisp: As I read this, Normalized is guaranteed to be a binary, where config:get/3 is guaranteed to be a list, so the first part of the orelse is guaranteed to be false.

iilyak:
Closed in favor of https://github.com/apache/couchdb-couch/pull/198

PR #63, opened 2015-06-30
Add a convenience macro for "equal to" test

kocolosk:
This macro preserves the expressiveness of ?assertEqual but uses == instead of =:=. Useful for situations where the test involves numbers that may be converted between types (e.g. 1 == 1.0).


rnewson:
neat. +1.

kxepal:
+1

PR #65, opened 2015-07-05
Force loopback if there are no admins

rnewson:
COUCHDB-2740


rnewson:
not sure this is the way to go.

PR #67, opened 2015-07-08
Throw bad request for empty POST request body when filtering by doc_ids

jaydoane:
BugzID: 48864

### DIFF
--- a/src/couch_changes.erl
+++ b/src/couch_changes.erl
@@ -309,6 +309,8 @@ get_view_qs(Req) ->
 
 get_doc_ids({json_req, {Props}}) ->
     check_docids(couch_util:get_value(<<"doc_ids">>, Props));
+get_doc_ids(#httpd{method='POST', req_body=undefined}) ->
+    throw({bad_request, "`_doc_ids` filter requires JSON body containing doc ids."});
 get_doc_ids(#httpd{method='POST'}=Req) ->
     {Props} = couch_httpd:json_body_obj(Req),
     check_docids(couch_util:get_value(<<"doc_ids">>, Props));
--- a/src/couch_changes.erl
+++ b/src/couch_changes.erl
@@ -310,7 +310,9 @@ get_view_qs(Req) ->
 get_doc_ids({json_req, {Props}}) ->
     check_docids(couch_util:get_value(<<"doc_ids">>, Props));
 get_doc_ids(#httpd{method='POST', req_body=undefined}) ->
-    throw({bad_request, "`_doc_ids` filter requires JSON body containing doc ids."});
+    throw({bad_request,
+           "`_doc_ids` filter requires JSON object body with field \"doc_ids\" "
+           "containing array of ids."});
 get_doc_ids(#httpd{method='POST'}=Req) ->
     {Props} = couch_httpd:json_body_obj(Req),
     check_docids(couch_util:get_value(<<"doc_ids">>, Props));


rnewson:
+1

kxepal:
This check is redundant since below check_docid does the same:

```
$curl -XPOST http://localhost:5984/db/_changes?filter=_doc_ids -d '' -H 'Content-Type:application/json'
{"error":"bad_request","reason":"invalid_json"}

$ curl -XPOST http://localhost:5984/db/_changes?filter=_doc_ids -d '{}' -H 'Content-Type:application/json'
{"error":"bad_request","reason":"`doc_ids` filter parameter is not a list."}
```

rnewson:
@kxepal this is for the case of no body, your example sends `{}`.

kxepal:
For no body it says invalid JSON as you see as well.

kxepal:
For bad payload there is another response:

```
$curl -XPOST http://localhost:5984/db/_changes?filter=_doc_ids -d '""' -H 'Content-Type:application/json'
{"error":"bad_request","reason":"Request body must be a JSON object"}
```

jaydoane:
But this handles the case when the request body is omitted entirely (not just an empty string).

rnewson:
@jaydoane @kxepal let's see these requests in full. Are they both Content-Length 0 or is there a difference?

kxepal:
@rnewson @jaydoane Full request/response for empty payload case (sorry for irrelevant examples):
```
*   Trying ::1...
* connect to ::1 port 5984 failed: Connection refused
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 5984 (#0)
> POST /db/_changes?filter=_doc_ids HTTP/1.1
> Host: localhost:5984
> User-Agent: curl/7.43.0
> Accept: */*
> Content-Type:application/json
> Content-Length: 0
> 
< HTTP/1.1 400 Bad Request
< Server: CouchDB/f29e590 (Erlang OTP/18)
< Date: Wed, 08 Jul 2015 21:14:33 GMT
< Content-Type: text/plain; charset=utf-8
< Content-Length: 48
< Cache-Control: must-revalidate
* HTTP error before end of send, stop sending
< 
{"error":"bad_request","reason":"invalid_json"}
* Closing connection 0
```

jaydoane:
```
curl -vX POST http://adm:pass@127.0.0.1:15984/db1/_changes?filter=_doc_ids -H "Content-Type: application/json"
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 15984 (#0)
* Server auth using Basic with user 'adm'
> POST /db1/_changes?filter=_doc_ids HTTP/1.1
> Authorization: Basic YWRtOnBhc3M=
> User-Agent: curl/7.37.1
> Host: 127.0.0.1:15984
> Accept: */*
> Content-Type: application/json
>
< HTTP/1.1 500 Internal Server Error
< X-CouchDB-Body-Time: 0
< X-Couch-Stack-Hash: 2380495304
< X-Couch-Request-ID: 4fc2df321e
* Server CouchDB/0bf8917 (Erlang OTP/17) is not blacklisted
< Server: CouchDB/0bf8917 (Erlang OTP/17)
< Date: Mon, 06 Jul 2015 06:38:47 GMT
< Content-Type: text/plain; charset=utf-8
< Content-Length: 59
< Cache-Control: must-revalidate
<
{"error":"function_clause","reason":null,"ref":2380495304}
```

kxepal:
@jaydoane ah, so this fix is for chttpd? This changes everything (thought it's strange that it miss the clause where couch_http doesn't).

kxepal:
+1 But may be make message a bit better?

```
`_doc_ids` filter requires JSON body containing doc ids." 
```

Implies that JSON body as array of doc ids is required, but that's not true. May be make it more precise like: 

```
`_doc_ids` filter requires JSON object body with field "doc_ids" containing array of ids.
```

Or throw the same invalid_json as couch_httpd does since empty payload is indeed invalid JSON.

rnewson:
notice that @jaydoane's request doesn't send Content-Length: 0, so that's why req_body is undefined rather than <<>>.

jaydoane:
@kxepal: no, I don't think this is specific for chttpd. It came up because someone had tried to use the _doc_ids changes filter without including a body and got a 500.

rnewson:
we just need a JIRA ticket now (and a force-push to this branch to update the commit).

jaydoane:
ok, I'll open a JIRA ticket, and can also change the wording. I like 
`_doc_ids` filter requires JSON object body with field "doc_ids" containing array of ids
if that's ok with @rnewson

kxepal:
@rnewson without Content-Length indeed there is function_clause error. Thanks! 
@jaydoane sorry for the confusion.

rnewson:
+1 to revised error message.

jaydoane:
https://issues.apache.org/jira/browse/COUCHDB-2743

rnewson:
ok, so squash these two commits and modify the comment message to mention `COUCHDB-2743` and then force push back to this branch.

jaydoane:
Thought I could accomplish same as force push by deleting remote branch and re-creating, but was wrong. There's no apparent way to re-open from this UI :(

rnewson:
I do it all the time, I promise you it's possible. force push from the command line.

jaydoane:
Yes, I understand that I should have done something like this:
git push origin 48864-bad-request-on-post-missing-body-doc-ids-filter --force
but I was lazy and just deleted the remote branch, which automatically closed this PR, and I the "Reopen" button is grayed out.

PR #69, opened 2015-07-11
Support Windows build target

wohali:


rebar.config.script
```diff
@@ -76,6 +77,9 @@ IcuDarwinEnv = [{"CFLAGS", "-DXP_UNIX -I/usr/local/opt/icu4c/include"},
                 {"LDFLAGS", "-L/usr/local/opt/icu4c/lib"}],
 IcuBsdEnv = [{"CFLAGS", "-DXP_UNIX -I/usr/local/include"},
              {"LDFLAGS", "-L/usr/local/lib"}],
+IcuWinEnv = [{"CFLAGS", "/DXP_WIN /IC:\\relax\\icu\\include"},
+             {"LDFLAGS", "/LIBPATH:C:\\relax\\icu\\lib64 icuin.lib icudt.lib icuuc.lib"}],
```
kxepal: Can we have "C:\relax" path configurable somehow with environment variable or else?

wohali: We could. Do you want that change before this PR lands?

kxepal: Would be good, but no urgency. I don't think we have a lot of people that will hit this path issue right after PR lands. So it's up to you.

kxepal:
+1

rebar.config.script
dch: given how much of a PITA it is to set up the toolchain I doubt it ;-).

It would be worth doing a vagrant image for this I think. /me makes a mental note.

dch: Hmm given this is in rebar.config.script can't we detect it somehow directly? I'm not exactly sure what we'd be looking for though.

rnewson:
+1

rebar.config.script
wohali: I didn't want to get too fancy. Env var to override the default location of where you unpacked ICU (I downloaded the prebuilt binaries, I didn't build ICU myself) is not a bad idea, though.

PR #70, opened 2015-07-12
add couch_util_clock:rfc1123()

robertkowalski:
as formatting for every request takes time, we cache the time
every second using a gen_server. cowboy has a very nice & clean
implementation of it, so i used this gen_server.

i was able to measure a 2% speed improvement on my laptop when
using it for obtaining the date of the response headers

COUCHDB-2734

### DIFF
--- a/src/couch_util_clock.erl
+++ b/src/couch_util_clock.erl
@@ -1,0 +1,217 @@
+%% Copyright (c) 2011-2014, Loïc Hoguin <essen@ninenines.eu>
+%%
+%% Permission to use, copy, modify, and/or distribute this software for any
+%% purpose with or without fee is hereby granted, provided that the above
+%% copyright notice and this permission notice appear in all copies.
+%%
+%% THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+%% WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+%% MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+%% ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+%% WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+%% ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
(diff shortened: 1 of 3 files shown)


rnewson:
hrm, a lot of code for a 2% improvement. have you looked at the new timing functions in erlang 18.0?

robertkowalski:
no not really, as we still support R14

robertkowalski:
but i measured that using erlangs `httpd_util:rfc1123_date/0` does not provide a real speed improvement

kxepal:
@rnewson in Erlang 18 time was reported to be much more slow then before:
http://erlang.org/pipermail/erlang-questions/2015-July/085081.html

src/couch_util_clock.erl
```diff
@@ -0,0 +1,217 @@
+%% Copyright (c) 2011-2014, Loïc Hoguin <essen@ninenines.eu>
```
kxepal: Since it's not our code, update of NOTICE file in apache/couchdb repository is required.

kxepal:
Indeed it's quite complicated solution for 2% while the idea here is pretty trivial: evaluate result only once and cache it and let everyone use that cache.

robertkowalski:
sorry for nit-picking, but obtaining the current time is not the performance problem here -- formatting it according to rfc1123 takes so much time. 

open http://robert-kowalski.de/assets/data/erlang-perf/chttpd-flame-server-header.svg and click on `mochiweb_request:server_headers/0` for an example. it even eats more time (percent-wise) on the backdoor ports: [here is a version that uses httpd_util:rfc1123 on the backdoor port](http://robert-kowalski.de/assets/data/2015-05-29-erlang-perf/flame--unpatched.svg)

i think 2% is OK, given that it adds up with the other performance improvements I already landed or that are in the work. I measured a 8% improvement for https://github.com/apache/couchdb-couch/commit/e19b262d4e5bd5ff1765b22589b8d3eeaac88749 yesterday on the cluster ports, so CouchDB is getting faster and faster by taking baby steps. I still hope for lager3 is released soon as I also reported some performance problems in lager to the lager team -- the fix for it should also bring ~5% better performance (btw the issue is also related to formatting of lists). taking all these three improvements we already have 15% which is a nice figure for relatively simple optimizations in the http layer.

rnewson:
I'm still against adding this complexity for a 2% difference, I don't think our attitude should be that we make things more complex than they already are for very little benefit.

I appreciate they can add up, but, as presented, this seems unnecessary.

janl:
I’m in favour of this, but I think this should live in mochiweb, not CouchDB.

kxepal:
@janl Good point, agree with this. Would be also good if this improvement will land into upstream.

rnewson:
moving the timestamp logic to mochiweb makes sense and you can therefore legitimately use mochiglobal to store the state ;)

PR #71, opened 2015-07-13
Optimize couch_ejson_compare NIF

davisp:
The old nif was allocating a set of collators that were reserved and
released by each scheduler thread. This coordination and the
accompanying mutex turned into a global point of contention when many
schedulers were using couch_ejson_compare.

This change removes the stack based concurrency control in favor of a
`__thread variable`. We end up with the same number of collators but
without any overhead around locking a central mutex.

This increases performance by roughly a factor of five.

COUCHDB-2732

### DIFF
--- a/priv/couch_ejson_compare/couch_ejson_compare.c
+++ b/priv/couch_ejson_compare/couch_ejson_compare.c
@@ -41,9 +41,10 @@ typedef struct {
     UCollator* coll;
 } ctx_t;
 
+static __thread UCollator* collator = NULL;
 static UCollator** collators = NULL;
-static int collStackTop = 0;
 static int numCollators = 0;
+static int numSchedulers = 0;
 static ErlNifMutex* collMutex = NULL;
 
 static ERL_NIF_TERM less_json_nif(ErlNifEnv*, int, const ERL_NIF_TERM []);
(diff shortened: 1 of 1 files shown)


kxepal:
+1 
Not sure how to test performance part for now. @robertkowalski do you have anything to fire here?

robertkowalski:
@davisp wowow! how are the numbers for a micro-benchmark that uses of one of our api-endpoints compared to an unpatched version?

I usually run something like this to compare patches:

```
#!/bin/sh

COUCHDB_PATH=$HOME/apache/couchdb
BENCH_RUN="siege -q -c120 -r100 -b http://127.0.0.1:15984/foo/testdoc"

sleep 60;
$COUCHDB_PATH/dev/run --with-admin-party-please & pid=$!
sleep 120

echo ""
echo ""
for i in `seq 1 10`;
do
  echo "Runnig test #$i:"
  $BENCH_RUN
  sleep 90
done

kill $pid

```

and reboot inbetween.  i think most important is that i have dropbox and goodle drive and all of their friends plus index services of the OS all disabled

davisp:
I'm working on getting the actual dataset used and benchmark code. The general outline of the test though is that they loaded something like 35K records into a Q=128 database from the XML music set and then had a view that emitted a row for the artist. The test was then to just run a view query getting 10 artists starting with a random letter between A-O (I assume there were less than 10 after O or something).

This test was then run with a number of different concurrent clients. IIRC somewhere around 100 clients maxed out throughput for both.

The exact numbers we were seeing were about 50-55 req/s with the current NIF implementation and about 340-350 req/s with this TLS version.

rnewson:
bump

kxepal:
Didn't test any performance, but for everything else it works good.
+1

PR #72, opened 2015-07-13
Erase async_open entry in failure case

rnewson:
COUCHDB-2749


kxepal:
+1

robertkowalski:
@rnewson can you add some context to the commit message so you don't have to search the issue in jira to understand the issue when reading the git log?

e.g.

```
Track open time without using process dictionary

We only remove the `{async_open, DbName}` process dictionary entries if 
the open is successful. Over time, in busy situations, the process dictionary
can grow very large.

```

robertkowalski:
it will also help us creating the next release log for 2.0 :)

PR #74, opened 2015-07-16
47059 session accept username

mayya-sharipova:


kxepal:
Hi! Please rebase your PR against master.

rnewson:
This also needs a JIRA ticket (and no reference to fogbugz). :)

rnewson:
@kxepal it's rebased now, what do you think? The goal is to accept "username" as an alias for "name" as cloudant's independently developed auth system expects that. I'd rather not maintain a fork of core couchdb for this, so I'd like it to go in.

To the patch itself then, it's the exact same `extract_username` function as in the original codebase that we imported to build couchdb 2.0, so it's a +1 from me.

src/couch_httpd_auth.erl
```diff
@@ -366,6 +366,18 @@ handle_session_req(#httpd{method='DELETE'}=Req, _AuthModule) ->
 handle_session_req(Req, _AuthModule) ->
     send_method_not_allowed(Req, "GET,HEAD,POST,DELETE").
 
+
+extract_username(Form) ->
+    CouchFormat = couch_util:get_value("name", Form),
+    try couch_util:get_value("username", Form, CouchFormat) of
```
kxepal: What's the case for having try/catch here? if `couch_util:get_value` will raise badarg due to `Form`, it'll happen on the line above.

src/couch_httpd_auth.erl
```diff
@@ -366,6 +366,18 @@ handle_session_req(#httpd{method='DELETE'}=Req, _AuthModule) ->
...
+    undefined ->
+      throw({bad_request, <<"user accounts must have a username">>});
```
kxepal: All around we use 4-spaces indention, not 2.

kxepal:
@rnewson I'm also +1 for this change, really helpful, except few moment I commented inline.

src/couch_httpd_auth.erl
```diff
@@ -366,6 +366,19 @@ handle_session_req(#httpd{method='DELETE'}=Req, _AuthModule) ->
 handle_session_req(Req, _AuthModule) ->
     send_method_not_allowed(Req, "GET,HEAD,POST,DELETE").
 
+extract_username(Form) ->
+    try couch_util:get_value("name", Form) of
```
kxepal: You miss the question about case of try/catch from previous response (:
There is no such case for this. To prove that let's take a look on L278-288 above.
`Form` expected to be a list, otherwise `couch_util:get_value` will raise `badarg` error. But, it couldn't be non-list.
If Content-Type is `application/x-www-form-urlencoded` then `mochiweb_util:parse_qs/1` called which returns a list.
If Content-Type is `application/json` then we match decoded json against `{Pairs}`. It payload is not a JSON object, we'll have `badmatch` here (btw, that's what is better to gracefully handle instead). Otherwise `Form` will be a list.
For every else Content-Type empty list is assignment. 

Any other case I miss when try/catch may be useful here?

src/couch_httpd_auth.erl
```diff
@@ -366,6 +366,19 @@ handle_session_req(#httpd{method='DELETE'}=Req, _AuthModule) ->
...
+    CouchFormat ->
```
kxepal: Not sure why not just `Username`...

src/couch_httpd_auth.erl
```diff
@@ -366,6 +366,19 @@ handle_session_req(#httpd{method='DELETE'}=Req, _AuthModule) ->
...
+        case couch_util:get_value("username", Form, CouchFormat) of
+        undefined ->
+            throw({bad_request, <<"user accounts must have a username">>});
```
kxepal: I believe, "request payload must contain a username" is more helpful error message, since user don't create any account here, but tries to log in.

src/couch_httpd_auth.erl
mayya-sharipova: Thanks for the detailed comments, I tried to address them.

rnewson:
Nice improvements but the term 'payload' is not right. Say 'request body' and we're good.

kxepal:
I'm bad with wording (: "request body" is good for me.

Logically, there is one more issue: if both name and username are provided, the latter will take the precedence. Might be good to detect that collision since we read both values in anyway, but I hardly could imagine the case when login form will have them both to POST to let this case hurt.

rnewson:
ah, sure, we should throw 400 Bad Request if you send name and username and they're not the same value.

mayya-sharipova:
Hopefully, the last commit addresses all comments :)

rnewson:
+1

kxepal:
+1

kxepal:
@mayya-sharipova thanks!

PR #75, opened 2015-07-17
Validate host header

rnewson:
By default, the Host header must be either localhost or the
fully-qualified domain name of the host.

Administrators can disable this check with;

[httpd]
validate_host = false

Administrators can override the valid host list with;

[httpd]
valid_hosts = foo,bar,baz

COUCHDB-2752

### DIFF
--- a/src/couch_httpd.erl
+++ b/src/couch_httpd.erl
@@ -30,6 +30,7 @@
 -export([send_json/2,send_json/3,send_json/4,last_chunk/1,parse_multipart_request/3]).
 -export([accepted_encodings/1,handle_request_int/5,validate_referer/1,validate_ctype/2]).
 -export([http_1_0_keep_alive/2]).
+-export([validate_host/1]).
 
 start_link() ->
     start_link(http).
@@ -293,6 +294,7 @@ handle_request_int(MochiReq, DefaultFun,
 
     {ok, Resp} =
     try
+        validate_host(HttpReq),
         check_request_uri_length(RawUri),
         case couch_httpd_cors:is_preflight_request(HttpReq) of
         #httpd{} ->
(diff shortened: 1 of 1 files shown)


kxepal:
It has a little bit flaw: assume `valid_hosts = foo,bar,,baz` - small typo after removing some host. Normal situation. Then valid hosts list will be ["foo", "bar", [], "baz"]. If request will be send with no Host header, then `hostname/1` will return empty list `[]` which will bypass validation.

rnewson:
I'll make it so Host header must exist if validate_hosts is true

kxepal:
Exists and be not empty, yes (:

rnewson:
If Host header is not sent, we look for, but can never find, the undefined atom in the list of valid hosts.

kxepal:
Ah, good then. Funny, but this PR breaks dev/run script - guess why (:

kxepal:
There CouchDB being requested by IP address 127.0.0.1 which is not in white default list.

src/couch_httpd.erl
```diff
@@ -357,6 +359,38 @@ handle_request_int(MochiReq, DefaultFun,
...
+            undefined;
+        Host ->
+            [Name | _Port] = re:split(Host, ":", [{parts, 2}, {return, list}]),
+            Name
+    end.
+
+valid_hosts() ->
+    case config:get("httpd", "valid_hosts") of
+        undefined ->
+            ["localhost", net_adm:localhost()];
```
kxepal: I think worth to add both "127.0.0.1" and "::1" here.

kxepal: Oh, no. IPv6 will break regexp above to extract hostname /:

rnewson:
@kxepal good catch on ip literals, I've updated. (though I note I use dev/run to test this and didn't hit a problem yet)

kxepal:
Awesome! +1

PR #76, opened 2015-07-18
Ensure doc groups are sorted before merging them

kocolosk:
We had been implicily assuming that clients send us sorted groups, but unsurprisingly that's not always the case. The additional sorting here should be redundant, but the consequences of merging unsorted groups are severe -- we can end up with uniqueness violations on the primary key in the database -- and so we add an additional sort here.

COUCHDB-2735


kxepal:
+1

kocolosk:
Fixing a typo in the commit message and merging, thanks for the review.

PR #77, opened 2015-07-19
Support Windows compilation for couch_ejson_compare

NorthNick:
MSVC uses __declspec(thread) rather than __thread, so created a macro for it.


rnewson:
+1

NorthNick:
Thanks.

davisp:
+1 even though I think this got merged this morning already.

NorthNick:
Thanks both.

Nick

PR #78, opened 2015-07-21
Don't enable host validation by default

rnewson:
Revising original stance. Administrators can enable this feature if
they want, and supply the definitive list of allowed Host values.

When enabled, the default list of valid hosts is empty.

COUCHDB-2752


kxepal:
+1

janl:
+1

PR #79, opened 2015-07-27
Avoid arithmetic in `to_hex` function

iilyak:


rnewson:
how much of a boost is this?

iilyak:
new         | old   
-------------|-------------
1096247 | 1489097 

Note: The number is a 99.9% for 10000 runs with 1000 bytes of data

kxepal:
+1

kxepal:
If you're interested, we played in to_hex benchmark two years ago. Summary could be found here: https://gist.github.com/maxlapshin/6809909
Matching with function clause is really the fastest method to do the things in Erlang.

robertkowalski:
wow thats quite cool! :)

rnewson:
+1

kocolosk:
+1 as well. Good use of the module constant pool and pattern matching to avoid the arithmetic here.

iilyak:
Thank you @kocolosk for the idea to use body recursion.

PR #80, opened 2015-07-31
Add CSRF protection

rnewson:
If the request parameter `csrf` is set to `true` when successfully
acquiring a session cookie from `_session` an additional cookie
(`Csrf-token`) is returned. All requests that send this new cookie
must also send a header (`X-Csrf-Token`) with the same value. If the
cookie is sent and the header is missing or different, a 403 response
is generated.

Note that the CSRF token is signed by the server so tampering is
detected and also results in a 403 response.

closes COUCHDB-2762

### DIFF
--- a/src/couch_httpd.erl
+++ b/src/couch_httpd.erl
@@ -295,6 +295,7 @@ handle_request_int(MochiReq, DefaultFun,
     {ok, Resp} =
     try
         validate_host(HttpReq),
+        couch_httpd_csrf:validate(HttpReq),
         check_request_uri_length(RawUri),
         case couch_httpd_cors:is_preflight_request(HttpReq) of
         #httpd{} ->
@@ -488,8 +489,9 @@ serve_file(#httpd{mochi_req=MochiReq}=Req, RelativePath, DocumentRoot,
     ResponseHeaders = server_header()
         ++ couch_httpd_auth:cookie_auth_header(Req, [])
         ++ ExtraHeaders,
-    {ok, MochiReq:serve_file(RelativePath, DocumentRoot,
-            couch_httpd_cors:cors_headers(Req, ResponseHeaders))}.
+    ResponseHeaders1 = couch_httpd_cors:cors_headers(Req, ResponseHeaders),
+    ResponseHeaders2 = couch_httpd_csrf:headers(Req, ResponseHeaders1),
+    {ok, MochiReq:serve_file(RelativePath, DocumentRoot, ResponseHeaders2)}.
 
 qs_value(Req, Key) ->
     qs_value(Req, Key, undefined).
@@ -661,7 +663,8 @@ start_response_length(#httpd{mochi_req=MochiReq}=Req, Code, Headers, Length) ->
     Headers1 = Headers ++ server_header() ++
                couch_httpd_auth:cookie_auth_header(Req, Headers),
     Headers2 = couch_httpd_cors:cors_headers(Req, Headers1),
-    Resp = MochiReq:start_response_length({Code, Headers2, Length}),
+    Headers3 = couch_httpd_csrf:headers(Req, Headers2),
+    Resp = MochiReq:start_response_length({Code, Headers3, Length}),
     case MochiReq:get(method) of
     'HEAD' -> throw({http_head_abort, Resp});
     _ -> ok
@@ -674,7 +677,8 @@ start_response(#httpd{mochi_req=MochiReq}=Req, Code, Headers) ->
     CookieHeader = couch_httpd_auth:cookie_auth_header(Req, Headers),
     Headers1 = Headers ++ server_header() ++ CookieHeader,
     Headers2 = couch_httpd_cors:cors_headers(Req, Headers1),
-    Resp = MochiReq:start_response({Code, Headers2}),
+    Headers3 = couch_httpd_csrf:headers(Req, Headers2),
+    Resp = MochiReq:start_response({Code, Headers3}),
     case MochiReq:get(method) of
         'HEAD' -> throw({http_head_abort, Resp});
         _ -> ok
@@ -709,7 +713,8 @@ start_chunked_response(#httpd{mochi_req=MochiReq}=Req, Code, Headers) ->
     Headers2 = Headers1 ++ server_header() ++
                couch_httpd_auth:cookie_auth_header(Req, Headers1),
     Headers3 = couch_httpd_cors:cors_headers(Req, Headers2),
-    Resp = MochiReq:respond({Code, Headers3, chunked}),
+    Headers4 = couch_httpd_csrf:headers(Req, Headers3),
+    Resp = MochiReq:respond({Code, Headers4, chunked}),
     case MochiReq:get(method) of
     'HEAD' -> throw({http_head_abort, Resp});
     _ -> ok
@@ -740,8 +745,9 @@ send_response(#httpd{mochi_req=MochiReq}=Req, Code, Headers, Body) ->
     Headers2 = Headers1 ++ server_header() ++
                couch_httpd_auth:cookie_auth_header(Req, Headers1),
     Headers3 = couch_httpd_cors:cors_headers(Req, Headers2),
+    Headers4 = couch_httpd_csrf:headers(Req, Headers3),
 
-    {ok, MochiReq:respond({Code, Headers3, Body})}.
+    {ok, MochiReq:respond({Code, Headers4, Body})}.
 
 send_method_not_allowed(Req, Methods) ->
     send_error(Req, 405, [{"Allow", Methods}], <<"method_not_allowed">>, ?l2b("Only " ++ Methods ++ " allowed")).
--- a/src/couch_httpd_csrf.erl
+++ b/src/couch_httpd_csrf.erl
@@ -1,0 +1,179 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License.  You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+%% This module provides optional CSRF protection to any client
+%%
+%% Clients should use the following pseudo code;
+%% if (hasCookie("CouchDB-CSRF")) {
+%%   setRequestHeader("X-CouchDB-CSRF", cookieValue("CouchDB-CSRF"));
+%% } else {
+%%   setRequestHeader("X-CouchDB-CSRF", "true")
+%% }
+%%
+%% If CouchDB sees the CouchDB-CSRF cookie then it checks its validity
+%% and whether the X-CouchDB-CSRF request header exists and matches.
+%% A 403 is returned if those checks fail.
+%% If CouchDB does not see the CouchDB-CSRF cookie but does see
+%% the X-CouchDB-CSRF header with value "true", a CouchDB-CSRF cookie
+%% is generated and returned.
+
+-module(couch_httpd_csrf).
+
+-export([validate/1, headers/2]).
+
+-include_lib("couch/include/couch_db.hrl").
+
+validate(#httpd{} = Req) ->
+    Cookie = csrf_from_req(Req),
+    Header = couch_httpd:header_value(Req, "X-CouchDB-CSRF"),
+    case {Cookie, Header} of
+        {undefined, undefined} ->
+            ok;
+        {undefined, "true"} ->
+            ok;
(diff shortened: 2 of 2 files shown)


robertkowalski:
it would be cool to have an integration test that proves that it works and protects against regressions

rnewson:
I was speaking with @kxepal and agreed to post the code so we could figure out _how_ to test it.

src/couch_httpd.erl
```diff
@@ -387,6 +388,35 @@ valid_hosts() ->
     List = config:get("httpd", "valid_hosts", ""),
     re:split(List, ",", [{return, list}]).
 
+validate_csrf(#httpd{} = Req) ->
+    case header_value(Req, "Cookie") of
+        undefined ->
+            ok;
+        Value ->
+            Cookies = mochiweb_cookies:parse_cookie(Value),
+            Cookie = couch_util:get_value("Csrf-token", Cookies),
```
kxepal: Can we have CSRF cookie name configurable to avoid possible collisions with other web services that also defined CSRF protection?  I cannot name any that uses the same "Csrf-token" value (Django have "csrftoken", Pyramid - "csrf_token"), but collision probability rate is still could be high here.

Or may be use "CouchDB-CSRF-Token" name by default - unlikely it will cause any problems.

rnewson: CouchDB-CSRF and X-CouchDB-CSRF sound fine to me.

src/couch_httpd_auth.erl
```diff
@@ -495,3 +508,11 @@ integer_to_binary(Int, Len) when is_integer(Int), is_integer(Len) ->
...
     Padded = <<Padding/binary, Unpadded/binary>>,
     binary:part(Padded, byte_size(Padded), -Len).
+
+csrf_cookie(Req) ->
+    Secret = ?l2b(ensure_csrf_secret()),
+    <<Token:8/binary, _/binary>> = couch_uuids:random(),
+    Hmac = crypto:sha_mac(Secret, Token),
+    mochiweb_cookies:cookie("Csrf-token",
+        couch_util:encodeBase64Url(<<Token/binary,":",Hmac/binary>>),
+        [{path, "/"}] ++ max_age()).
```
kxepal: Isn't a whole point of CSRF to reissue a token after each request or after some period?

rnewson: Some CSRF mitigation patterns do that but it's not the only method. We're doing this;

https://www.owasp.org/index.php/Cross-Site_Request_Forgery_%28CSRF%29_Prevention_Cheat_Sheet#Double_Submit_Cookies

src/couch_httpd_auth.erl
```diff
@@ -304,12 +310,19 @@ handle_session_req(#httpd{method='POST', mochi_req=MochiReq}=Req, AuthModule) ->
             UserSalt = couch_util:get_value(<<"salt">>, UserProps2),
             CurrentTime = make_cookie_time(),
             Cookie = cookie_auth_cookie(Req, ?b2l(UserName), <<Secret/binary, UserSalt/binary>>, CurrentTime),
+            % maybe set the csrf token
+            Cookies = case couch_httpd:qs_value(Req, "csrf", "false") of
```
kxepal: mmm...so we here
1. Enable CSRF protection only by user request, not for the whole server. What means, that it will be easy to forget to activate that protection.
2. CSRF is not applied for Basic and OAuth authed users and those that authed using other methods as they don't POST /_session. They could only GET it, but it's not mandatory. 

What are your thoughts about these cases?

src/couch_httpd_auth.erl
kxepal: Good to know about other methods. Thanks for sharing!

src/couch_httpd_auth.erl
rnewson: If we don't make it optional then every client will be broken until they add this custom header.

On basic auth, I tend to think of cookie auth being used by browsers/ui's and basic auth being used by programs. CSRF is only an issue for browsers; direct api interactions aren't vulnerable.

rnewson: I meant https://en.wikipedia.org/wiki/Cross-site_request_forgery#Cookie-to-Header_Token earlier, sorry for confusion.

rnewson:
if we now understand the general mechanism, my remaining questions are;

1) unless the admin sets allow_persistent_cookies the x-couchdb-csrf cookie is only lost when the browser is restarted or cookies are manually cleared
2) if allow_persistent_cookies is true, the x-couchdb-csrf cookie will expire. the auth session cookie will expire at the same time, unless it's been renewed (which happens automatically), so this would seem to lock the user out eventually.

The core parts are sound, I think. The generation of the csrf cookie and the way we validate that the cookie and header match and are valid. What's not quite right is the way you get the csrf cookie in the first place.

Thoughts?

rnewson:
I've revised this and removed the coupling with the session cookie. 

A client that does not currently have a CouchDB-CSRF cookie can add X-CouchDB-CSRF: true to their request and will receive one. They must then send the cookie's value in the X-CouchDB-CSRF header for their request to succeed.

The CSRF token check occurs before authentication so you can acquire the CSRF token immediately (perhaps when fetching the welcome message from /).

The token will expire from time to time (same duration as the session cookie but without the automatic extension), so clients should do broadly this;

if (hasCookie("CouchDB-CSRF")) {
  setRequestHeader("X-CouchDB-CSRF", cookieValue("CouchDB-CSRF");
} else {
  setRequestHeader("X-CouchDB-CSRF", "true");
}

kxepal:
I like new logic much more. Good improvements!

However, your questions are still valid: if CSRF cookie expires, how it will get prolonged? Especially, since auth cookie renews automagicaly when it ttl comes to the end.

rnewson:
Because the client will be testing if they have a CSRF cookie (in order to know whether to set the CSRF header). It will vanish, so they will set X-CouchDB-CSRF:true on the next request to get a new one.

kxepal:
I see. This could work.

rnewson:
hm, I'm thinking of adding automatic refresh, like we do for session cookie, to prevent even that little gap where there's no CSRF token at expiration.

kxepal:
I think such refresh could be useful and making CSRF tokens expire is even better.

rnewson:
So I've added the refresh.When a cookie is more than halfway through its lifetime (which defaults to an hour) a replacement cookie is sent.

Additionally, if you send the CSRF header but not the cookie, that's a 400 Bad Request. This helps clients stay safe (they won't think that just sending the header and getting a non-error response means that they are protected by the CSRF mechanism.

rnewson:
The list of tests we need;

1) a request with no cookie and with `X-CouchDB-CSRF` header set to true results in `Set-Cookie` response header for a cookie named `CouchDB-CSRF`.
2) a request that has a `CouchDB-CSRF` cookie but not `X-CouchDB-CSRF` request header is rejected with a 403.
3) a request that has a `CouchDB-CSRF` cookie and a `X-CouchDB-CSRF` request header, but that do not match each other, is rejected with a 403.
4) a request that has a `CouchDB-CSRF` cookie and a `X-CouchDB-CSRF` request header, that do match each other, is accepted.
5) a request that has an expired `CouchDB-CSRF cookie and a `X-CouchDB-CSRF` request header, that do match each other, is rejected with a 400.
5) a request that has no `CouchDB-CSRF cookie and a `X-CouchDB-CSRF` request header, that do match each other, is rejected with a 400.
6) a request that has a `CouchDB-CSRF` cookie and a `X-CouchDB-CSRF` request header, that do match each other, made more than halfway through the cookie's lifetime, is accepted and a new  `CouchDB-CSRF` is returned (with a full lifetime).

src/couch_httpd_csrf.erl
```diff
@@ -0,0 +1,147 @@
...
+        undefined ->
+            NewSecret = ?b2l(couch_uuids:random()),
+            config:set("couch_httpd_auth", "csrf_secret", NewSecret),
+            NewSecret;
+        Secret -> Secret
+    end.
+
+
+decode_cookie(Cookie) ->
+    Cookie1 = couch_util:decodeBase64Url(Cookie),
```
kxepal: How about to catch a case if we receive malformed cookie here? While HTTP 500 because of badmatch or badarg is Erlang way, but more user friendly will be to be HTTP 400 here.

rnewson: sure.

rnewson:
Another small enhancement, if the cookie is invalid (corrupt or if the secret value was changed) then we clear the cookie, allowing the client to recover without waiting for expiration (which might be 59 minutes away!)

src/couch_httpd_csrf.erl
```diff
@@ -0,0 +1,177 @@
...
+        {undefined, "true"} ->
+            ok;
+        {undefined, _} ->
+            throw({forbidden, <<"CSRF header sent without cookie">>});
+        {Csrf, Csrf} ->
+            %% cookie and header match, but is it valid?
+            ok = validate(Csrf),
+
+            %% check expiration
+            {_, Timestamp, _} = decode_cookie(Csrf),
```
kxepal: You do `decode_cookie/1` in `validate/1` above where you handle `invalid` result, but here you don't. Sure, it's not logically possible to get badmatch here, but may be move check expiration to validate function to avoid decoding same cookie twice?

kxepal:
Sorry for double comment posting. GH eventually put my comments to your fork commit, not PR. Not sure how this happens. Also, check code for tabs: you have space-tabs mix a bit.

rnewson:
turns out a lot of kits don't delete cookie if you set an expiration date in the past, so instead we send a fresh cookie if we get sent an invalid one.

src/couch_httpd_csrf.erl
```diff
@@ -0,0 +1,172 @@
...
+
+headers(#httpd{} = Req, Headers) ->
+    Header = couch_httpd:header_value(Req, "X-CouchDB-CSRF"),
+    case {csrf_from_req(Req), csrf_in_headers(Headers), Header} of
+        {undefined, false, "true"} ->
+            [make_cookie() | Headers];
+        {Csrf, false, Csrf} when Csrf /= undefined ->
+            case decode_cookie(Csrf) of
+                invalid ->
+                    [make_cookie() | Headers];
```
kxepal: Is there a case for this branch? `validate/1` calls in `handle_request_int` which will throw `bad_request` if `Csrf` is invalid while I see `headers/1` is only called in response when `Csrf` may be `undefined` or valid cookie as otherwise it will not pass validation. Or I miss something?

rnewson:
Ok, I've stopped editing now, was just tidying a few things up.

kxepal:
Perfect! Like latest changes about malformed and config section change. Have only notes about styling is you don't mind.

src/couch_httpd_csrf.erl
```diff
@@ -0,0 +1,170 @@
...
+    Header = couch_httpd:header_value(Req, "X-CouchDB-CSRF"),
+    case {Cookie, Header} of
+        {undefined, undefined} ->
+            ok;
+        {undefined, "true"} ->
+            ok;
+        {"deleted", "true"} ->
+            ok;
+        {undefined, _} ->
+            throw({forbidden, <<"CSRF header sent without cookie">>});
```
kxepal: It's the only place where header and cookie words not capitalized.

src/couch_httpd_csrf.erl
```diff
@@ -0,0 +1,170 @@
...
+                                    [make_cookie() | Headers];
+                                false ->
+                                    Headers
+                            end;
+                        false ->
+                            [delete_cookie() | Headers]
+                    end
+            end;
+        _ ->
+	    Headers
```
kxepal: Tab is used.

src/couch_httpd_csrf.erl
```diff
@@ -0,0 +1,170 @@
...
+    mochiweb_cookies:cookie("CouchDB-CSRF", "deleted",
+        [{path, "/"}, {max_age, 0}]).
+
+csrf_from_req(#httpd{} = Req) ->
+    case couch_httpd:header_value(Req, "Cookie") of
+	undefined ->
+	    undefined;
+	Value ->
+	    Cookies = mochiweb_cookies:parse_cookie(Value),
+	    couch_util:get_value("CouchDB-CSRF", Cookies)
```
kxepal: Tabs is used in 113-117 lines.

src/couch_httpd_csrf.erl
```diff
@@ -0,0 +1,170 @@
...
+decode_cookie(Cookie) ->
+    try
+        Cookie1 = couch_util:decodeBase64Url(Cookie),
+        <<Token:8/binary, Timestamp:32, Hmac:20/binary>> = Cookie1,
+        {Token, Timestamp, Hmac}
+    catch
+        error:_ ->
+            malformed
+    end.
+
```
kxepal: Missing newline.

src/couch_httpd_csrf.erl
rnewson: bah stupid editor

rnewson:
@kxepal +1?

kxepal:
@rnewson +1!

PR #81, opened 2015-07-31
Epi hooks

iilyak:
This is one of the series of PRs to make couchdb plugable.
This PR depends on:
- https://github.com/apache/couchdb-chttpd/pull/53
- https://github.com/apache/couchdb-couch-index/pull/7
- https://github.com/apache/couchdb-couch-mrview/pull/23
- https://github.com/apache/couchdb-couch-replicator/pull/14
- https://github.com/apache/couchdb-fabric/pull/24
- https://github.com/apache/couchdb-global-changes/pull/8

### DIFF
--- a/include/couch_db.hrl
+++ b/include/couch_db.hrl
@@ -92,8 +92,16 @@
     auth,
     default_fun,
     url_handlers,
-    authentication_handlers = []
+    authentication_handlers = [],
+    absolute_uri,
+    auth_module,
+    begin_ts,
+    original_method,
+    nonce,
+    cors_config,
+    qs
 }).
+
 
 -record(doc, {
     id = <<"">>,
--- a/include/couch_db.hrl
+++ b/include/couch_db.hrl
@@ -41,12 +41,12 @@
 -define(ADMIN_CTX, {user_ctx, ?ADMIN_USER}).
 
 -define(SYSTEM_DATABASES, [
-    "_dbs",
-    "_global_changes",
-    "_metadata",
-    "_nodes",
-    "_replicator",
-    "_users"
+    <<"_dbs">>,
+    <<"_global_changes">>,
+    <<"_metadata">>,
+    <<"_nodes">>,
+    <<"_replicator">>,
+    <<"_users">>
 ]).
 
 
--- a/include/couch_db.hrl
+++ b/include/couch_db.hrl
@@ -91,9 +91,9 @@
     design_url_handlers,
     auth,
     default_fun,
-    url_handlers
+    url_handlers,
+    authentication_handlers = []
 }).
-
 
 -record(doc, {
     id = <<"">>,
--- a/src/couch_db.erl
+++ b/src/couch_db.erl
@@ -34,10 +34,14 @@
 -export([check_md5/2, with_stream/3]).
 -export([monitored_by/1]).
 -export([normalize_dbname/1]).
+-export([validate_dbname/1]).
 
 -include_lib("couch/include/couch_db.hrl").
 
--define(VALID_DB_NAME, "^[a-z][a-z0-9\\_\\$()\\+\\-\\/]*$").
+-define(DBNAME_REGEX,
+    "^[a-z][a-z0-9\\_\\$()\\+\\-\\/]*" % use the stock CouchDB regex
+    "(\\.[0-9]{10,})?$" % but allow an optional shard timestamp at the end
+).
 
 start_link(DbName, Filepath, Options) ->
     case open_db_file(Filepath, Options) of
@@ -1480,3 +1484,32 @@ normalize_dbname(<<"shards/", _/binary>> = Path) ->
     lists:last(binary:split(mem3:dbname(Path), <<"/">>, [global]));
 normalize_dbname(DbName) ->
     DbName.
+
+validate_dbname(DbName) when is_list(DbName) ->
+    validate_dbname(?l2b(DbName));
+validate_dbname(DbName) when is_binary(DbName) ->
+    Normalized = normalize_dbname(DbName),
+    case couch_db_plugin:validate_dbname(DbName, Normalized) of
+        true ->
+            ok;
+        false ->
+            validate_dbname_int(DbName, Normalized)
+    end.
+
+validate_dbname_int(DbName, Normalized) when is_binary(DbName) ->
+    case re:run(DbName, ?DBNAME_REGEX, [{capture,none}, dollar_endonly]) of
+        match ->
+            ok;
+        nomatch ->
+            case is_systemdb(Normalized) of
+                true -> ok;
+                false -> {error, {illegal_database_name, DbName}}
+            end
+    end.
+
+is_systemdb(DbName) when is_list(DbName) ->
+    is_systemdb(?l2b(DbName));
+is_systemdb(<<"shards/", _/binary>> = Path) when is_binary(Path) ->
+    is_systemdb(normalize_dbname(Path));
+is_systemdb(DbName) when is_binary(DbName) ->
+    lists:member(?b2l(DbName), ?SYSTEM_DATABASES).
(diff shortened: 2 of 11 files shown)


src/couch_db_plugin.erl
```diff
@@ -0,0 +1,65 @@
...
+]).
+
+-include_lib("couch/include/couch_db.hrl").
+
+%% ------------------------------------------------------------------
+%% API Function Definitions
+%% ------------------------------------------------------------------
+
+validate_dbname(DbName, Normalized) ->
+    Handle = couch_epi:get_handle(couch_db),
```
kxepal: In other apps, you defined `?SERVICE_ID` macro for handle name. Why not here as well?

include/couch_db.hrl
```diff
@@ -91,10 +91,27 @@
...
+    absolute_uri,
+    auth_module,
+    begin_ts,
+    original_method,
+    nonce,
+    cors_config,
+    qs
+}).
+
+-record(httpd_res, {
```
kxepal: Shouldn't this record and changes about be in chttpd module?

kxepal:
I'm quite confused by multiple DbName conversion from list to binary and back. Can we always use DbName as binary and just fix error message generation and `?SYSTEM_DATABASES`? Should simplify some code a bit.

src/couch_httpd.erl
```diff
@@ -872,7 +872,7 @@ error_info({bad_ctype, Reason}) ->
     {415, <<"bad_content_type">>, Reason};
 error_info(requested_range_not_satisfiable) ->
     {416, <<"requested_range_not_satisfiable">>, <<"Requested range not satisfiable">>};
-error_info({error, illegal_database_name, Name}) ->
+error_info({error, {illegal_database_name, Name}}) ->
```
kxepal: I guess if I provide my own validate_dbname hook with different regexp constraint, this error message will be true no more, right?

src/couch_httpd_auth.erl
```diff
@@ -346,10 +346,10 @@ handle_session_req(#httpd{method='GET', user_ctx=UserCtx}=Req, _AuthModule) ->
                 ]}},
                 {info, {[
                     {authentication_db, ?l2b(config:get("couch_httpd_auth", "authentication_db"))},
-                    {authentication_handlers, [auth_name(H) || H <- couch_httpd:make_fun_spec_strs(
-                            config:get("httpd", "authentication_handlers"))]}
```
kxepal: It's not possible any more to enable disabled by default auth handlers like [`proxy_authentication_handler`](https://github.com/apache/couchdb-couch/blob/master/src/couch_httpd_auth.erl#L118-L138) or [`null_authentication_handler`](https://github.com/apache/couchdb-couch/blob/master/src/couch_httpd_auth.erl#L115-L116). Also, JavaScript test suite is broken by impossibility to set [`
special_test_authentication_handler`](https://github.com/apache/couchdb-couch/blob/master/src/couch_httpd_auth.erl#L33-L52).

include/couch_db.hrl
iilyak: I can move it. However this would mean that the plugin developer would need to do:
```
-include_lib("chttpd/include/chttpd.hrl").
```
Which is alright.

src/couch_httpd.erl
iilyak: This is hard to accomplish. Because of [this](https://github.com/cloudant/couchdb-couch/blob/epi_hooks/src/couch_db_plugin.erl#L32). `couch_epi:any` would succeed when any of the plugins would return `true`. So there is no way to inject custom error message. Since for next plugin the same dbname could be valid. I can see two ways to do custom messages. None of them perfect.
1. `throw({illegal_database_name, DbName, Msg})` and handle this error in `chttpd_plugin:handle_error` handler in your plugin. 
2. Re-implement couch_db_plugin:validate_dbname to return list of `{true, ""} | {false, Error}`. If there are no `true` value in the list concatenate error messages. 
3. Implement `couch_db_plugin:invalid_database_message/2` hook.

src/couch_httpd_auth.erl
iilyak: The way `authentication_handlers` were implemented is inconsistent anyway. The chttpd app has them hardcoded. It also doesn't set the value for `UserCtx#user_ctx.handler`. So `authenticated` field is completely missing if called on clustered interface. I'll think about it more. What would be your suggestions? Any hints?

kxepal: Well, quite naive one: move these to own micro-plugins (which could be still live in this repo) and provide config options to enable/disable these handlers directly.

As for filling UserCtx fields, may be just backport that logic from httpd? Sounds not harmful for the first sight.

include/couch_db.hrl
kxepal: I think that ok if end user goes to extend HTTP layer.

src/couch_httpd.erl
kxepal: I think the first option is the most sane and correct - validate function is only holds the knowledge about database name restrictions and throwing error with descriptive message  from there is correct. chttpd need just transform this error into HTTP response.

iilyak:
@davisp: I fixed most of you comments. The one which is left is about the typo in a commit message. I'll fix it on squash.

davisp:
Mostly minor comments from me on this and the related PRs. The overall structure and bulk of the work looks great. Just a few weird things that look like churn from development that didn't get flattened out before making the PR and then some possibly unrelated stuff that may have just been found during this work (which I'm fine with as long as there's a good commit message to explain what is going on with some of the subtler changes).

include/couch_db.hrl
```diff
@@ -91,10 +91,17 @@
...
+    absolute_uri,
+    auth_module,
+    begin_ts,
+    original_method,
+    nonce,
+    cors_config,
+    qs
 }).
 
-
```
rnewson: double spacing was intentional.

src/couch_db.erl
```diff
@@ -1480,3 +1497,32 @@ normalize_dbname(<<"shards/", _/binary>> = Path) ->
...
+validate_dbname(DbName) when is_binary(DbName) ->
+    Normalized = normalize_dbname(DbName),
+    case couch_db_plugin:validate_dbname(DbName, Normalized) of
+        true ->
+            ok;
+        false ->
+            validate_dbname_int(DbName, Normalized)
+    end.
+
+validate_dbname_int(DbName, Normalized) ->
```
rnewson: an is_binary(DbName) here would be a useful check, it's guaranteed by `validate_dbname` above.

src/couch_httpd.erl
```diff
@@ -161,8 +161,15 @@ set_auth_handlers() ->
     AuthenticationSrcs = make_fun_spec_strs(
         config:get("httpd", "authentication_handlers", "")),
     AuthHandlers = lists:map(
-        fun(A) -> {make_arity_1_fun(A), ?l2b(A)} end, AuthenticationSrcs),
-    ok = application:set_env(couch, auth_handlers, AuthHandlers).
+        fun(A) -> {auth_handler_name(A), make_arity_1_fun(A)} end, AuthenticationSrcs),
+    AuthenticationFuns = AuthHandlers ++ [
+        {<<"local">>, fun couch_httpd_auth:local_admin_handler/1} %% should be last
+    ],
+    ok = application:set_env(couch, auth_handlers, AuthenticationFuns).
```
rnewson: hrm, this is unpleasant, application env is not persistent, are we using here as a cache?

rnewson: ah, but I guess is predates your work. See my comment on the name. summary: party_mode_handler would be more descriptive (and suitably scary)

src/couch_httpd.erl
```diff
@@ -161,8 +161,15 @@ set_auth_handlers() ->
...
+
+auth_handler_name(SpecStr) ->
+    [_,_,_,_,_,Name|_] = re:split(SpecStr, "[\\W_]", []),
```
rnewson: `lists:nth` would be clearer.

rnewson:
good stuff, some comments to address from myself and kxepal. will re-review after.

include/couch_db.hrl
iilyak: should I double space all records or it will be a change non related to PR?

rnewson:
+1

PR #82, opened 2015-08-01
correct inverted test function name

alanning:
Based on this [article](http://robert-kowalski.de/blog/lets-learn-erlang-and-fix-a-bug-on-a-couchdb-cluster/), the function is ensuring that a username is returned.  This change just removes the "not" from the test function name.

### DIFF
--- a/test/couchdb_auth_tests.erl
+++ b/test/couchdb_auth_tests.erl
@@ -34,7 +34,7 @@ auth_test_() ->
                 foreach,
                 fun setup/0, fun teardown/1,
                 [
-                    fun should_not_return_username_on_post_to_session/1
+                    fun should_return_username_on_post_to_session/1
                 ]
             }
         }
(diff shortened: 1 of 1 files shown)


robertkowalski:
thanks for the patch! :)

did the testsuite work for you? I think https://github.com/alanning/couchdb-couch/blob/ca9af59663fdc42f53d0850e4b1bc1e4c4e6dff5/test/couchdb_auth_tests.erl#L37 also needs a change (list of tests that are going to run)

kxepal:
Agree with @robertkowalski - test suite will fail because of unknown test function.

alanning:
Ah, thanks!  I made the change purely through github's interface and didn't realize there was a separate list of tests.  Travis CI was showing "pending..." when I created the PR; I should have let it run to completion.

alanning:
Hmm, TravisCI tests still failing but it seems unrelated. "No Rakefile found".

kxepal:
Ignore travis by now. LGFM.

PR #83, opened 2015-08-08
Add option to make CSRF cookie mandatory

rnewson:
When enabled, the CSRF cookie/header is required on all requests
except those to the welcome message at /.

COUCHDB-2762

### DIFF
--- a/src/couch_httpd_csrf.erl
+++ b/src/couch_httpd_csrf.erl
@@ -37,11 +37,11 @@ validate(#httpd{} = Req) ->
     Header = couch_httpd:header_value(Req, "X-CouchDB-CSRF"),
     case {Cookie, Header} of
         {undefined, undefined} ->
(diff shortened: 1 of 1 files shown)


src/couch_httpd_csrf.erl
```diff
@@ -63,6 +63,16 @@ validate(Csrf) when is_list(Csrf) ->
             end
     end.
 
+throw_if_mandatory(#httpd{path_parts = []}) ->
+    ok; %% Welcome message is public / entrypoint
```
kxepal: Here could be not welcome message, but database endpoint, thanks to vhosts and rewrite feature.

kxepal:
+1

kxepal:
Tested, all fine, sorry for confusion. Too many paths, too many servers (: +1, it works as expected.

PR #84, opened 2015-08-10
CSRF is not necessary for read operations

rnewson:
COUCHDB-2762

### DIFF
--- a/src/couch_httpd_csrf.erl
+++ b/src/couch_httpd_csrf.erl
@@ -32,6 +32,12 @@
 
 -include_lib("couch/include/couch_db.hrl").
 
+validate(#httpd{method = 'GET'}) ->
+    ok;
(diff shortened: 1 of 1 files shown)


kxepal:
+1

I see only problems with _update handler(where GET is disabled on query server side, not httpd => custom query server may create/modify documents on GET) and custom externals which could not respect REST and do modifications on GET, but here we can nothing much to do.

rnewson:
Any preference for a single clause version?

`validate(#httpd{method = Method}) when Method == 'GET'; Method == 'HEAD'; Method == 'OPTIONS' ->`

kxepal:
@rnewson if you fit it in 80 line right margin - ok (: Personally, I found your initial version more readable - less syntax noise around the valuable data, but fine with both solutions.

rnewson:
I'll go with the original then. thanks!

PR #85, opened 2015-08-12
Implement rename_on_delete

iilyak:
Sometimes it is usefull to keep deleted db files around. For backup
reasons for example. This commit implements the feature. The feature is
enabled by specifying `rename_on_delete = true` in `[couchdb]` of the
config.

### DIFF
--- a/src/couch_server.erl
+++ b/src/couch_server.erl
@@ -460,9 +460,7 @@ handle_call({delete, DbName, Options}, _From, Server) ->
 
         couch_db_plugin:on_delete(DbName, Options),
 
-        Async = not lists:member(sync, Options),
-
-        case couch_file:delete(Server#server.root_dir, FullFilepath, Async) of
+        case delete_db_file(Server#server.root_dir, FullFilepath, Options) of
         ok ->
             couch_event:notify(DbName, deleted),
             {reply, ok, Server2};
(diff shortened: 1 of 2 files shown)


davisp:
+1

kxepal:
`couch_file:delete` already does rename + move into special directory. May be we just need to extend it in order to not actually delete these files?

rnewson:
but it renames it to a random file, the point here is we can move it back (undelete)

rnewson:
+1 from me

kxepal:
nobody prevents to let it rename with not random name (:

rnewson:
hm, fair point. we could consolidate.

rnewson:
ah, but no, we also leave it in the original directory, we'd have to encode the path into the new name to make this work

rnewson:
+1 as-is.

kxepal:
Even if so, I believe renaming function still have to be in couch_file module, not in couch_server. Because there we have everything to work with files and directories.

rnewson:
and the other point is that .delete is wiped at startup, and the value of rename_on_delete is that nothing in couchdb deletes the files, it's done externally.

kxepal:
Good point about wipe. I forgot about it.

rnewson:
@kxepal +1 then?

kxepal:
So what about to move rename to `couch_file`? I searched over code base, we almost everywhere trying to use couch_file for FS operations. Exceptions are PID file write, old compaction file delete on init, but it's looks like a mistake and other `file:rename` calls which could be rerouted to `couch_file` module after.

But I can and happy to refactor this afterwards and then +1 this PR.

rnewson:
I think couch_server is the right place. couch_file handles a particular file, I don't think it should be involved in renaming (we call file:delete not couch_file:delete).

kxepal:
Fine, I was though that `couch_file` had a bit different role then. +1

rnewson:
looking back, couch_file was the better spot. This change doesn't provide this rename instead of delete option for compaction.

iilyak:
@kxepal @rnewson: I don't like calling `config` from `couch_file`.

kxepal:
@iilyak Call config in couch_server and pass the value to couch_file for further execution? Sounds like a compromise.

iilyak:
In this case it needs to be done in other places as well. Including the places (`couch_mrview`) which are not suppose to know anything about rename feature:
- `couch/src/couch_db_updater.erl`
- `couch/src/couch_server.erl`
- `couch_mrview/src/couch_mrview_compactor.erl`
- `couch_mrview/src/couch_mrview_util.erl`

iilyak:
Unless we use couch_server for all calls from other apps.

iilyak:
But then it is the same as updating `couch_db_updater.erl` and `couch/src/couch_server.erl` to call `couch_server`.

iilyak:
I'll figure something out

kxepal:
@iilyak :+1:

PR #86, opened 2015-08-12
Return bad_request on view compilation error

mikewallace1979:
A request adding a non-compiling view was previously generating
an HTTP 500 response. However, because the request is supplying an
invalid view an HTTP 400 bad request response would be appropriate
here. It would also make things clearer to the client that the
fault lies within the request rather than the server.

This commit updates the error returned on view compilation error
to a {bad_request, compilation_error, Msg}. This will be converted
into an HTTP 400 response by chttpd.

Closes COUCHDB-2772

### DIFF
--- a/src/couch_query_servers.erl
+++ b/src/couch_query_servers.erl
@@ -41,7 +41,7 @@ try_compile(Proc, FunctionType, FunctionName, FunctionSource) ->
     catch {compilation_error, E} ->
         Fmt = "Compilation of the ~s function in the '~s' view failed: ~s",
         Msg = io_lib:format(Fmt, [FunctionType, FunctionName, E]),
-        throw({compilation_error, Msg})
+        throw({bad_request, compilation_error, Msg})
     end.
 
 start_doc_map(Lang, Functions, Lib) ->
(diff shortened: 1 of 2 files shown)


src/couch_query_servers.erl
```diff
@@ -41,7 +41,7 @@ try_compile(Proc, FunctionType, FunctionName, FunctionSource) ->
     catch {compilation_error, E} ->
         Fmt = "Compilation of the ~s function in the '~s' view failed: ~s",
         Msg = io_lib:format(Fmt, [FunctionType, FunctionName, E]),
-        throw({compilation_error, Msg})
+        throw({bad_request, compilation_error, Msg})
```
kxepal: I think you should change `couch_db:validate_ddoc` instead. Because this function may be called for the valid user request while complication will fail and then CouchDB will lie again.

mikewallace1979:
@kxepal I made the change in `couch_db:validate_ddoc` instead. Interestingly I had to wrap the validate call in a case because the validation function converts throws into tuples and returns them [1].

That being the case should we also move the `invalid_design_doc` catch on L583 into a case clause?

[1] https://github.com/apache/couchdb-couch-index/blob/master/src/couch_index_server.erl#L59-L60

kxepal:
@mikewallace1979 Interesting. If `couch_index_server:validate/2` turns all the errors into retvals, then additional try/catch in `couch_db:validate_ddoc/2` is useless (almost). I think yes, better replace that try/catch with case then.

@rnewson what you say?

kxepal:
cc @davisp as well as he implemented this logic in 22fdbe2 and apache/couchdb-couch-index@fd046ec6 . May be we miss something tricky here.

kxepal:
See https://github.com/apache/couchdb-couch/pull/97

mikewallace1979:
Superseded by #97

PR #88, opened 2015-08-18
Temporarily lock out authentication attempts after repeated failures

rnewson:
COUCHDB-2778

### DIFF
--- a/src/couch_auth_delay.erl
+++ b/src/couch_auth_delay.erl
@@ -1,0 +1,189 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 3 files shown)


kxepal:
```$ git apply 88.patch
88.patch:209: new blank line at EOF.
+
warning: 1 line adds whitespace errors.
```

kxepal:
Behavior is strange: when lock happens CouchDB doesn't sends any response back holding the connection. So I believe, this could be used to cause resources leak on server.

rnewson:
for a lockout to be effective it has to slow the attacker, the only way to do that is to make requests take longer. That does tie up the socket, sadly. Any thoughts?

kxepal:
@rnewson I think, much better to just log it and configure fail2ban or similar system which will generate firewall rules about. TARPIT for iptables can cause a lot of troubles for attacker and it's much more effective that we can implement. So far, there is no need to implement poor man firewall with which we can shoot ourselfs.

kxepal:
Also, it seems locking doesn't happens if I provide non existed username. So with current behaviour I can figure which users are registered during bruteforce.

rnewson:
two good points, will ponder.

rnewson:
I'll rethink, thanks for the feedback.

PR #89, opened 2015-08-21
Bind both to IPv4 and IPv6 with special "any" value

kxepal:
This also removes need in special wrapper to handle atom as default
for config:get call.

COUCHDB-2783

### DIFF
--- a/src/couch_httpd.erl
+++ b/src/couch_httpd.erl
@@ -31,6 +31,7 @@
 -export([accepted_encodings/1,handle_request_int/5,validate_referer/1,validate_ctype/2]).
 -export([http_1_0_keep_alive/2]).
 -export([validate_host/1]).
+-export([validate_bind_address/1]).
 
 -define(HANDLER_NAME_IN_MODULE_POS, 6).
 
@@ -90,8 +91,11 @@ start_link(https) ->
          {ssl_opts, SslOpts}],
     start_link(https, Options).
 start_link(Name, Options) ->
-    BindAddress = with_default(config:get("httpd", "bind_address"), any),
-    validate_bind_address(BindAddress),
+    BindAddress = case config:get("httpd", "bind_address", "any") of
+                      "any" -> any;
+                      Else -> Else
+                  end,
+    ok = validate_bind_address(BindAddress),
     DefaultSpec = "{couch_httpd_db, handle_request}",
     DefaultFun = make_arity_1_fun(
         config:get("httpd", "default_handler", DefaultSpec)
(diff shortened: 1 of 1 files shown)


rnewson:
This only helps for the private 5986 port for 2.0 which should really only bind to localhost. Look at https://github.com/apache/couchdb-chttpd/blob/master/src/chttpd.erl#L112.

kxepal:
@rnewson I know. See also https://github.com/apache/couchdb-chttpd/pull/60

rnewson:
+1

rnewson:
hrm, I retract the +1, sorry.

src/couch_httpd.erl
```diff
@@ -1127,11 +1127,10 @@ check_for_last(#mp{buffer=Buffer, data_fun=DataFun}=Mp) ->
                 data_fun = DataFun2})
     end.
 
+validate_bind_address(any) -> any;
+validate_bind_address("any") -> any;
```
rnewson: I don't think the validate_bind_address function should work this way, it should _only_ validate. It now returns 'ok' or 'any' or throws, which is quite confusing. The intention is correct though, to allow config to hold "any" and have that be the atom any. Have another go :)

kxepal: Right (:

kxepal:
@rnewson  Fixed that problem. Please take another look.

src/couch_httpd.erl
```diff
@@ -1127,11 +1131,9 @@ check_for_last(#mp{buffer=Buffer, data_fun=DataFun}=Mp) ->
                 data_fun = DataFun2})
     end.
 
+validate_bind_address(any) -> ok;
 validate_bind_address(Address) ->
     case inet_parse:address(Address) of
-        {ok, _} -> ok;
```
rnewson: change this back to `ok`, you never use the return value anyway.

src/couch_httpd.erl
```diff
@@ -88,7 +89,10 @@ start_link(https) ->
          {ssl_opts, SslOpts}],
     start_link(https, Options).
 start_link(Name, Options) ->
-    BindAddress = with_default(config:get("httpd", "bind_address"), any),
+    BindAddress = case config:get("httpd", "bind_address", "any") of
+                      "any" -> any;
+                      Else -> Else
+                  end,
     validate_bind_address(BindAddress),
```
rnewson: say `ok = ` here.

src/couch_httpd.erl
kxepal: Accidental change. Thanks.

rnewson:
+1, please squash before merging.

PR #91, opened 2015-08-24
Fix crashing filtered _changes request

eiri:
A filtered call to _changes end-point prompts a node to
pull a peer from the received mochi request record.
This is crashing on the secondary nodes in a clustered
environment since the peer call made on a socket
that is not original for them.

The fix makes use of the fact that by the time of the request
the peer already got resolved on the coordinator node and
passed along as a part of #http record.

This closes the case COUCHDB-2785


rnewson:
+1

PR #93, opened 2015-08-27
Capture histogram data for _bulk_requests

rnewson:
COUCHDB-2792

### DIFF
--- a/priv/stats_descriptions.cfg
+++ b/priv/stats_descriptions.cfg
@@ -50,6 +50,10 @@
     {type, counter},
     {desc, <<"number of _local document write operations">>}
 ]}.
+{[couchdb, httpd, bulk_docs], [
+    {type, histogram},
+    {desc, <<"distribution of the number of docs in _bulk_docs requests">>}
+]}.
 {[couchdb, httpd, bulk_requests], [
     {type, counter},
     {desc, <<"number of bulk requests">>}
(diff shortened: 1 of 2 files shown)


priv/stats_descriptions.cfg
```diff
@@ -51,8 +51,8 @@
     {desc, <<"number of _local document write operations">>}
 ]}.
 {[couchdb, httpd, bulk_requests], [
-    {type, counter},
-    {desc, <<"number of bulk requests">>}
+    {type, histogram},
+    {desc, <<"distribution of bulk request sizes">>}
```
kxepal: I think this metric should be renamed. Now it's not about how much bulk requests were made, but how many documents were passed in bulk. And request size here confusing: without reading the code I would say that this is amount of bytes passed with request.

Suggestion: `[couchdb, httpd, bulk_docs]` and `distribution of amount docs stored in bulk`.

Sounds good?

rnewson: Yeah, I'm stuck on names. The 'n' value in here is the same value as the original counter, which is why I felt the name was appropriate, it just has more information, like the distribution of the sizes of these requests. That you can have multiple things inside a _bulk_docs is what distinguishes it from most other endpoints, so I felt that implied what the histogram meant (and the description makes it explicit)

kxepal: Let's keep counter in additional to histogram. This shouldn't hurt, but there nobody would be confused about which field to use. Also, `n` here wouldn't be the same as old counter in case of invalid request.

kxepal:
+1

PR #95, opened 2015-09-02
check POST requests for valid json header

mayya-sharipova:
validate that all POST requests with json body must have also have valid
json header: {"Content-Type": "application/json"}
This ensures a basic protection against CSRF

JIRA: COUCHDB-2775

### DIFF
--- a/src/couch_httpd_misc_handlers.erl
+++ b/src/couch_httpd_misc_handlers.erl
@@ -185,6 +185,8 @@ handle_config_req(#httpd{method='GET', path_parts=[_, Section, Key]}=Req) ->
     end;
 % POST /_config/_reload - Flushes unpersisted config values from RAM
 handle_config_req(#httpd{method='POST', path_parts=[_, <<"_reload">>]}=Req) ->
+    couch_httpd:validate_ctype(Req, "application/json"),
+    _ = couch_httpd:body(Req),
     ok = couch_httpd:verify_is_server_admin(Req),
     ok = config:reload(),
     send_json(Req, 200, {[{ok, true}]});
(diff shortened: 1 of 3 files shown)


src/couch_httpd_misc_handlers.erl
```diff
@@ -185,6 +185,7 @@ handle_config_req(#httpd{method='GET', path_parts=[_, Section, Key]}=Req) ->
     end;
 % POST /_config/_reload - Flushes unpersisted config values from RAM
 handle_config_req(#httpd{method='POST', path_parts=[_, <<"_reload">>]}=Req) ->
+    couch_httpd:validate_ctype(Req, "application/json"),
```
rnewson: this one doesn't take a request body.

rnewson:
+1 after fixing that one superfluous check.

src/couch_httpd_misc_handlers.erl
kxepal: _compact and _ensure_full_commit requests doesn't take any body as well, but they also ensures in JSON mimetype. So it's for consistent behaviour. However, need to forcely read the body `_ = couch_httpd:body(Req),` to prevent request hangs.

mayya-sharipova: Thanks for comments. For consistency reason, I left validate_ctype in _reload request, also reading the body:  _ = couch_httpd:body(Req).

kxepal: Oh, right, sorry, didn't notice it.

rnewson: hm, ok then.

rnewson:
+1

PR #99, opened 2015-09-11
Allow filtering of active tasks by property

rnewson:
COUCHDB-2807

### DIFF
--- a/src/couch_task_status.erl
+++ b/src/couch_task_status.erl
@@ -28,6 +28,7 @@
 -export([start_link/0, stop/0]).
 -export([all/0, add_task/1, update/1, get/1, set_update_frequency/1]).
 -export([is_task_added/0]).
+-export([filter/2]).
 
 -export([init/1, terminate/2, code_change/3]).
 -export([handle_call/3, handle_cast/2, handle_info/2]).
@@ -47,6 +48,10 @@ stop() ->
 
 all() ->
     gen_server:call(?MODULE, all).
+
+
+filter(Key, Value) ->
+    gen_server:call(?MODULE, {filter, Key, Value}).
 
 
 add_task(Props) ->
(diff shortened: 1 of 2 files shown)


src/couch_task_status.erl
```diff
@@ -122,6 +127,14 @@ handle_call({add_task, TaskProps}, {From, _}, Server) ->
     [_] ->
         {reply, {add_task_error, already_registered}, Server}
     end;
+handle_call({filter, Key, Value}, _, Server) ->
+    All = [
```
iilyak: Use following to avoid doing the same thing in a tight loop
```
handle_call({filter, Key0, Value0}, _, Server) ->
Key = couch_util:to_list(Key0),
Value = couch_util:to_list(Value0),
```

src/couch_task_status.erl
```diff
@@ -130,6 +143,12 @@ handle_call(all, _, Server) ->
     ],
     {reply, All, Server}.
 
+filter_task(Key, Value, TaskProps) ->
```
rnewson: not a huge fan of this function but I don't see much option

src/couch_task_status.erl
```diff
@@ -122,6 +127,14 @@ handle_call({add_task, TaskProps}, {From, _}, Server) ->
     [_] ->
         {reply, {add_task_error, already_registered}, Server}
     end;
+handle_call({filter, Key, Value}, _, Server) ->
+    All = [
+        [{pid, ?l2b(pid_to_list(Pid))} | TaskProps]
+        ||
+        {Pid, TaskProps} <- ets:tab2list(?MODULE),
```
iilyak: Would it be possible to use `ets:foldl(Function, [], ?MODULE)` here?

rnewson:
hm, do we like this?

kxepal:
I do like idea, just didn't test yet. Btw, how it will correlate with [COUCHDB-2405](https://issues.apache.org/jira/browse/COUCHDB-2405)?

PR #103, opened 2015-09-23
Fix crypto deprecations

rnewson:
COUCHDB-2825


kxepal:
+1 here and for others.

P.S. You miss couch-epi, I handled it: https://github.com/apache/couchdb-couch-epi/pull/11

rnewson:
I deliberately skipped it as it calls erlang:md5 which is not deprecated and to avoid making it depend on couchdb.

kxepal:
Ah, ok. I'll close it then.

kxepal:
However, it still emits crypto warning. I changed PR to handle it in different way.

PR #105, opened 2015-09-28
Inherit io_priority from the process creating the stream

tsutsu:
`couch_stream` should be able to reuse the `io_priority` of whatever process creates it; it's then the creator's responsibility to decide what `io_priority` the stream should have.

Assigning streaming IO operations their own separately-tunable category of `io_priority` is an alternative to this change. The usefulness of that would depend on whether streaming IO in general has predictable load impacts that would be useful to observe and control for. Since `couch_stream` is only currently used by `couch_att`, though, giving attachment upload operations specifically an `io_priority` category might make more sense.

COUCHDB-2828
BugzID: 25815

### DIFF
--- a/src/couch_stream.erl
+++ b/src/couch_stream.erl
@@ -51,7 +51,7 @@ open(Fd) ->
     open(Fd, []).
 
 open(Fd, Options) ->
-    gen_server:start_link(couch_stream, {Fd, self(), Options}, []).
+    gen_server:start_link(couch_stream, {Fd, self(), erlang:get(io_priority), Options}, []).
 
 close(Pid) ->
     gen_server:call(Pid, close, infinity).
@@ -198,7 +198,8 @@ write(Pid, Bin) ->
     gen_server:call(Pid, {write, Bin}, infinity).
 
 
-init({Fd, OpenerPid, Options}) ->
+init({Fd, OpenerPid, OpenerPriority, Options}) ->
+    erlang:put(io_priority, OpenerPriority),
     {EncodingFun, EndEncodingFun} =
     case couch_util:get_value(encoding, Options, identity) of
     identity ->


src/couch_stream.erl
```diff
@@ -51,7 +51,7 @@ open(Fd) ->
     open(Fd, []).
 
 open(Fd, Options) ->
-    gen_server:start_link(couch_stream, {Fd, self(), Options}, []).
+    gen_server:start_link(couch_stream, {Fd, {self(), erlang:get(io_priority)}, Options}, []).
```
rnewson: why the extra nesting?

tsutsu: It's a bit of semantic idealism. Since `self()` is already being passed, and the gen_server could theoretically extract the `io_priority` from the passed pid() later on (using `erlang:process_info(Pid, dictionary)`, though it's inefficient and bad form), you could say that self() "contains" the io_priority in an OOP-theoretic sense. Effectively, the result of using `erlang:get/1` is *derived data* of `self()`, equivalent to something like the computed length of a list.

Frequently, you want to pass an object around along with its derived data, in such a way that its consumers think of the composite as an ADT with the same functionality as the original object. In a class-based language, you'd do something silly here with a decorator proxy to enable the object you're passing to respond to a `get_io_priority` message, but otherwise behave the same. In Erlang, you'd either use a record (if you want to create a contract on the shape of the ADT), or simply stick the original data together with its derived data into a tuple.

...pragmatically, though, that's a bunch of hot air and it's an extra structure/destructure op and could just as well be eliminated.

(It *would* have more practical value if any intermediate function calls occurred that had just expected "an owner", where the composite `{OwnerPid, OwnerPriority}` could "ride through" as an abstract `Owner` and then be unwrapped only by functions that care. But none do.)

rnewson: the init function would crash either way, so I think it's clearer if each required argument is expressed naturally.

tsutsu:
Nesting removed.

rnewson:
+1

PR #107, opened 2015-09-29
Add revs limit for docs passed to filter functions

chewbranca:
The existing revs_limit logic only accounts for depth of individual
branches, and ignores the case where you have a lot of branches, for
instance when a doc is heavily conflicted. This patch truncates the list
of revs when passing docs to the JS filter functions, because given
enough revs this can cause the couchjs processes to explode.

### DIFF
--- a/src/couch_query_servers.erl
+++ b/src/couch_query_servers.erl
@@ -280,12 +280,24 @@ validate_doc_update(DDoc, EditDoc, DiskDoc, Ctx, SecObj) ->
             throw({unknown_error, Message})
     end.
 
-json_doc(nil) -> null;
+json_doc_options() ->
+    json_doc_options([]).
+
+json_doc_options(Options) ->
+    Limit = config:get_integer("query_server_config", "revs_limit", 20),
+    [{revs, Limit} | Options].
+
 json_doc(Doc) ->
-    couch_doc:to_json_obj(Doc, [revs]).
+    json_doc(Doc, json_doc_options()).
+
+json_doc(nil, _) ->
+    null;
+json_doc(Doc, Options) ->
+    couch_doc:to_json_obj(Doc, Options).
 
 filter_view(DDoc, VName, Docs) ->
-    JsonDocs = [couch_doc:to_json_obj(Doc, [revs]) || Doc <- Docs],
+    Options = json_doc_options(),
+    JsonDocs = [json_doc(Doc, Options) || Doc <- Docs],
     [true, Passes] = ddoc_prompt(DDoc, [<<"views">>, VName, <<"map">>], [JsonDocs]),
     {ok, Passes}.
 
@@ -296,7 +308,8 @@ filter_docs(Req, Db, DDoc, FName, Docs) ->
     #httpd{} = HttpReq ->
         couch_httpd_external:json_req_obj(HttpReq, Db)
     end,
-    JsonDocs = [couch_doc:to_json_obj(Doc, [revs]) || Doc <- Docs],
+    Options = json_doc_options(),
+    JsonDocs = [json_doc(Doc, Options) || Doc <- Docs],
     [true, Passes] = ddoc_prompt(DDoc, [<<"filters">>, FName],
         [JsonDocs, JsonReq]),
     {ok, Passes}.
--- a/src/couch_query_servers.erl
+++ b/src/couch_query_servers.erl
@@ -268,14 +268,13 @@ get_number(Key, Props) ->
 validate_doc_update(DDoc, EditDoc, DiskDoc, Ctx, SecObj) ->
     JsonEditDoc = couch_doc:to_json_obj(EditDoc, [revs]),
     JsonDiskDoc = json_doc(DiskDoc),
-    Resp = ddoc_prompt(DDoc,
-                       [<<"validate_doc_update">>],
-                       [JsonEditDoc, JsonDiskDoc, Ctx, SecObj]),
-    case Resp of
-        _ when Resp /= 1 ->
-            couch_stats:increment_counter(
-                [couchdb, query_server, vdu_rejects], 1);
-        _ -> ok
+    Resp = ddoc_prompt(
+        DDoc,
+        [<<"validate_doc_update">>],
+        [JsonEditDoc, JsonDiskDoc, Ctx, SecObj]
+    ),
+    if Resp == 1 -> ok; true ->
+        couch_stats:increment_counter([couchdb, query_server, vdu_rejects], 1)
     end,
     case Resp of
         1 ->
@@ -311,10 +310,10 @@ filter_view(DDoc, VName, Docs) ->
 
 filter_docs(Req, Db, DDoc, FName, Docs) ->
     JsonReq = case Req of
-    {json_req, JsonObj} ->
-        JsonObj;
-    #httpd{} = HttpReq ->
-        couch_httpd_external:json_req_obj(HttpReq, Db)
+        {json_req, JsonObj} ->
+            JsonObj;
+        #httpd{} = HttpReq ->
+            couch_httpd_external:json_req_obj(HttpReq, Db)
     end,
     Options = json_doc_options(),
     JsonDocs = [json_doc(Doc, Options) || Doc <- Docs],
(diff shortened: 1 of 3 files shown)


rnewson:
hrm, when filtering is there any point passing the _rev or rev history at all? (cc @davisp @kocolosk).

kxepal:
@chewbranca Does this also affects on other design functions? IIRC shows also receives documents with the whole _revisions list by default now. Not sure about update ones, never was curious about.

Another question is why filter function receives list of _revisions now? What the use case of this and may be it's better to not send it at all?

davisp:
@rnewson @kxepal When I first saw this my first thought was to remove _revisions entirely assuming that it was basically an accident to have included it in the first place. But given that its there and its theoretically possible someone might rely on it I instead suggested the truncation approach since revs_limit exists so anyone relying on it can't correctly rely on a specific length of the _revisions list.

kxepal:
@davisp I'm not sure about someone to relay on it. What's practical use case for them in filter function? String all docs which has more than N revisions in history? Sounds strange.

rnewson:
hrm, fair enough, we can't be sure no one is depending on it. 

+1

davisp:
@kxepal I haven't got a clue why someone might use it. But it exists so I'm guessing someone somewhere would be angry if we removed it.

kxepal:
@davisp well, good point. Let's land this and run some poll later to deprecate and remove this. 

However, we also need more general solution since here we improved filtered changes feed performance a lot, but still I can blow your query server with show and update function using the same vector. Having `filter_revs_limit`, `shows_revs_limit`, `update_revs_limit` config options is a bit weird.

src/couch_query_servers.erl
```diff
@@ -296,7 +296,8 @@ filter_docs(Req, Db, DDoc, FName, Docs) ->
     #httpd{} = HttpReq ->
         couch_httpd_external:json_req_obj(HttpReq, Db)
     end,
-    JsonDocs = [couch_doc:to_json_obj(Doc, [revs]) || Doc <- Docs],
+    Limit = config:get_integer("couchdb", "filter_revs_limit", 20),
```
kxepal: May be use 
```
Limit = config:get_integer("query_servers", "revs_limit", 20),
```

? Then we can reuse the same option for other design functions which receives _revisions list now.

chewbranca: I went with `couchdb` because `revs_limit` is part of the `db_header` record, but as this is for options going through the query server, that switch to `query_server_config` seems reasonable to me.

chewbranca: Fixed in 1628a3a.

src/couch_query_servers.erl
```diff
@@ -296,7 +296,8 @@ filter_docs(Req, Db, DDoc, FName, Docs) ->
     #httpd{} = HttpReq ->
         couch_httpd_external:json_req_obj(HttpReq, Db)
     end,
-    JsonDocs = [couch_doc:to_json_obj(Doc, [revs]) || Doc <- Docs],
+    Limit = config:get_integer("query_server_config", "filter_revs_limit", 20),
```
kxepal: You still think it only applicable for filter functions? How about this:
```
--- a/src/couch_query_servers.erl
+++ b/src/couch_query_servers.erl
@@ -282,7 +282,14 @@ validate_doc_update(DDoc, EditDoc, DiskDoc, Ctx, SecObj) ->
 
 json_doc(nil) -> null;
 json_doc(Doc) ->
-    couch_doc:to_json_obj(Doc, [revs]).
+    json_doc(Doc, json_doc_options()).
+json_doc(Doc, Options) ->
+    couch_doc:to_json_obj(Doc, Options).
+
+json_doc_options() ->
+    Limit = config:get_integer("query_server_config", "revs_limit", 20),
+    Options = [{revs, Limit}],
+    Options.
 
 filter_view(DDoc, VName, Docs) ->
     JsonDocs = [couch_doc:to_json_obj(Doc, [revs]) || Doc <- Docs],
@@ -296,7 +303,8 @@ filter_docs(Req, Db, DDoc, FName, Docs) ->
     #httpd{} = HttpReq ->
         couch_httpd_external:json_req_obj(HttpReq, Db)
     end,
-    JsonDocs = [couch_doc:to_json_obj(Doc, [revs]) || Doc <- Docs],
+    Options = json_doc_options(),
+    JsonDocs = [json_doc(Doc, Options) || Doc <- Docs],
     [true, Passes] = ddoc_prompt(DDoc, [<<"filters">>, FName],
         [JsonDocs, JsonReq]),
     {ok, Passes}.
```

Then this option will get shared for [shows](https://github.com/apache/couchdb-couch-mrview/blob/master/src/couch_mrview_show.erl#L78) and [updates](https://github.com/apache/couchdb-couch-mrview/blob/master/src/couch_mrview_show.erl#L123) functions.

chewbranca: @kxepal I looked at all the places that call `couch_query_server:json_doc/1` and I think you're right, that would be a good safe option to making the `revs_limit` logic work here for show, filter, and update functions. Any objections @davisp @rnewson?

davisp: Nope.

rnewson:
+1. @chewbranca merge pls

kxepal:
@rnewson comment wasn't addressed.

rnewson:
which comment?

src/couch_query_servers.erl
```diff
@@ -280,12 +287,23 @@ validate_doc_update(DDoc, EditDoc, DiskDoc, Ctx, SecObj) ->
...
+json_doc_options() ->
+    json_doc_options([]).
+
+json_doc_options(Options) ->
+    Limit = config:get_integer("query_server_config", "revs_limit", 20),
+    [{revs, Limit} | Options].
+
 json_doc(Doc) ->
-    couch_doc:to_json_obj(Doc, [revs]).
+    json_doc(Doc, json_doc_options()).
```
kxepal: It was good idea to keep `json_doc(nil) -> null;` to not make config lookup when we don't actually need in it result.

src/couch_query_servers.erl
```diff
@@ -280,12 +287,23 @@ validate_doc_update(DDoc, EditDoc, DiskDoc, Ctx, SecObj) ->
...
+
+json_doc(nil, _) ->
+    null;
+json_doc(Doc, Options) ->
+    couch_doc:to_json_obj(Doc, Options).
 
 filter_view(DDoc, VName, Docs) ->
-    JsonDocs = [couch_doc:to_json_obj(Doc, [revs]) || Doc <- Docs],
+    JsonDocs = [json_doc(Doc) || Doc <- Docs],
```
kxepal: Here is better explicitly pass options because otherwise you'll query config for N times as much as Docs you have.

chewbranca: Sure, fixed this and the other place doing the same thing in 9c5b8b7.

src/couch_query_servers.erl
chewbranca: I disagree, I took it out intentionally. I'm not a fan of duplicating that logic; the one arity `json_doc` function now exists to add default options, and should do nothing else. The two arity version handles all the actual logic. I did consider the impact of adding the extra config lookup before short circuiting on `nil`, but I don't think we should be worrying about the speed of config lookups from ETS in general, especially at the expense of clarity in the code.

kxepal: Ok, you have a point. Thanks!

kxepal:
+1

rnewson:
after appropriate squashing

chewbranca:
Merged in master.

kxepal:
Merged with https://github.com/apache/couchdb-couch/commit/eb60e41ad0afdad1281455e19cf1bf02419864d4 .

rnewson:
@chewbranca how come this isn't marked 'merged' like usual?

chewbranca:
@rnewson I forgot to push back out the branch after I rebased my changes on top of master, so the sha's were different in master than here.

kxepal:
@rnewson Commit hashes in PR and on master are different. And merge commit doesn't contains magic works "This closes #107". One of these conditions need to be fixed to help ASFBot get to know which PR need to close/merge.

rnewson:
@chewbranca @kxepal it's certainly possible to do this 'right', can we be more careful for future PR's please.

kxepal:
@rnewsown Sure. Actually just need to always push changes to your feature branch before do the actual merge. If we'll ask for / do rebase against current master before all of that - would be awesome and this help us to avoid useless merge commits for single commit changes.

PR #109, opened 2015-09-30
Fix OS Daemons test dependency.

nickva:
In isolation was failing in configuration_reader_test_
Because fixtures/os_daemon_configer.escript was
requesting uuids.algorithm from config application,
which was not set up.

COUCHDB-2831

### DIFF
--- a/test/couchdb_os_daemons_tests.erl
+++ b/test/couchdb_os_daemons_tests.erl
@@ -41,6 +41,9 @@ setup(DName) ->
     {ok, OsDPid} = couch_os_daemons:start_link(),
     config:set("os_daemons", DName,
                      filename:join([?FIXTURESDIR, DName]), false),
+    % Set configuration option to be used by configuration_reader_test_
+    % This will be used in os_daemon_configer.escript:test_get_cfg2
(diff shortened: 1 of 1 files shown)


test/couchdb_os_daemons_tests.erl
```diff
@@ -41,6 +41,7 @@ setup(DName) ->
     {ok, OsDPid} = couch_os_daemons:start_link(),
     config:set("os_daemons", DName,
                      filename:join([?FIXTURESDIR, DName]), false),
+    config:set("uuids","algorithm","sequential", false),
```
kxepal: Spaces after commas and add notice about why we need it here

kxepal:
Previously, this config option was read from default.ini files that was loaded by config app. It seems now it's gone and config loads with empty state. While it's good to restore old behavior, it also shows how much our tests implicitly coupled with other things and these dependencies are not very clear.

So it's ok to explicitly define dependency from this uuid setting here, since we don't test default algorithm here, but ability of os_daemon to read retrieve some config setting via stdio commands.

+1 after address comment.

PR #110, opened 2015-09-30
Fix task status unit test dependency

nickva:
It depends on couch_log to be running. Otherwise it
throw this error:

```
unexpected termination of test process*
::{{badmatch,undefined},
   [{couch_log,debug,2,[{file,"src/couch_log.erl"},{line,32}]},
    {couch_task_status,handle_cast,2,
                       [{file,"src/couch_task_status.erl"},{line,137}]},
```

And only 6 out of 11 tests are run.

COUCHDB-2832

### DIFF
--- a/test/couch_task_status_tests.erl
+++ b/test/couch_task_status_tests.erl
@@ -19,14 +19,18 @@
 
 
 setup() ->
+    Ctx = test_util:start(?MODULE, [couch_log], [{dont_mock, [config]}]),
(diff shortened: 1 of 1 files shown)


kxepal:
+1 OOM issue is gone too

kxepal:
Though, I'm scare about the behavior we have for one missed dep. Don't we hide real monsters under bed?

PR #112, opened 2015-09-30
Remove round for create/delete database for 1000 times

kxepal:
I'm not sure what we testing here except our disk drives healthiness.


kxepal:
@rnewson do you remember why we had this?

rnewson:
nope, except to flush out race conditions. +1

PR #113, opened 2015-10-01
Fix "initialization from incompatible pointer type" warning for icu drv

kxepal:
It was caused by using ErlDrvSSizeT type instead of ErlDrvSizeT. It
was not an issue prior to R15 when ErlDrvSSizeT and ErlDrvSizeT where
basically int's.

Since we raised minimal Erlang release requirement up to R16B03-1,
there is no reason to maintain compatibility with pre-R15 for driver
so local COUCH_* types are replaced with ErlDrv* ones.


rnewson:
+1

PR #115, opened 2015-10-06
Export delete file

iilyak:
In order to support `rename_on_delete` semantics for views. We want to reuse `couch_server:delete_file` in couch_mrview_cleanup.

### DIFF
--- a/src/couch_server.erl
+++ b/src/couch_server.erl
@@ -25,6 +25,8 @@
 % config_listener api
 -export([handle_config_change/5, handle_config_terminate/3]).
 
+-export([delete_file/3]).
+
 -include_lib("couch/include/couch_db.hrl").
 
 -define(MAX_DBS_OPEN, 100).
@@ -460,8 +462,8 @@ handle_call({delete, DbName, Options}, _From, Server) ->
 
         couch_db_plugin:on_delete(DbName, Options),
 
-        case delete_db_file(Server#server.root_dir, FullFilepath, Options) of
-        ok ->
+        case delete_file(Server#server.root_dir, FullFilepath, Options) of
+        {ok, _} ->
             couch_event:notify(DbName, deleted),
             {reply, ok, Server2};
         {error, enoent} ->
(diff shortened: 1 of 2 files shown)


rnewson:
+1, and nice cleanup of a cut-and-paste in the test. The assumption that the rename takes less than one second predates your change but it would be nice to get rid of it.

src/couch_server.erl
```diff
@@ -549,9 +554,12 @@ delete_db_file(RootDir, FullFilePath, Options) ->
     end.
 
 rename_on_delete(Original) ->
+    file:rename(Original, renamed_filename(Original)).
+
+renamed_filename(Original) ->
```
kxepal: It's actually "deleted_filename", unless you make `deleted` suffix customizable.

src/couch_server.erl
```diff
@@ -549,9 +554,13 @@ delete_db_file(RootDir, FullFilePath, Options) ->
     end.
 
 rename_on_delete(Original) ->
+    file:rename(Original, deleted_filename(Original)).
+
+deleted_filename(Original) ->
     {{Y,Mon,D}, {H,Min,S}} = calendar:universal_time(),
+    UUID = ?b2l(couch_uuids:random()),
```
rnewson: I get what you intended but this seems awkward. Could delete_file instead return `ok | {renamed, NewName} | {error, Reason}` instead? Then we know where to look in the tests.

iilyak: Good idea. I'll change it to return:
- `{ok, deleted}`
- `{ok, {renamed, FileName}}`
- `{error, Reason}`

src/couch_server.erl
```diff
@@ -538,20 +540,31 @@ db_closed(Server, Options) ->
...
 rename_on_delete(Original) ->
+    DeletedFileName = deleted_filename(Original),
+    case file:rename(Original, DeletedFileName) of
+        ok -> {ok, {renamed, DeletedFileName}};
+        Else -> Else
+    end.
+
+deleted_filename(Original) ->
     {{Y,Mon,D}, {H,Min,S}} = calendar:universal_time(),
+    UUID = ?b2l(couch_uuids:random()),
```
rnewson: no need for the uuid, timestamp suffices

rnewson:
+1 after reverting the uuid change. the original names were fine for uniqueness, the issue was the assumption that the test runs within one second (it would fail spuriously if the clock ticked between some steps).

iilyak:
I'll revert the UUID change and squash for final review.

rnewson:
+1

PR #119, opened 2015-10-13
Stabilize couch_task_status_tests:run_multiple_tasks/0 test.

nickva:
Previously, at the end of the test, it kills each status process
with call(Pid1, done), then immediately calls couch_task_status:all().

This caused a race between couch_task_status receiving the 'all'
call and the 'DOWN' message (from monitoring test processes). If
'DOWN' message came before 'all', test passed. If it came
after, it failed.

### DIFF
--- a/test/couch_task_status_tests.erl
+++ b/test/couch_task_status_tests.erl
@@ -180,6 +180,16 @@ loop() ->
             From ! {ok, self(), ok}
     end.
 
+call(Pid, done) ->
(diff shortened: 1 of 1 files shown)


kxepal:
Can we instead of sleep, add start the monitor for Pid on call done and also hang there for some time till DOWN? Should be more generic solution which is not limited by 100ms.

kxepal:
+1

PR #120, opened 2015-10-14
COUCHDB-769: Store attachments in the external storage. 

gilv:
Initial implementation that allows CouchDB to store attachments outside of the database file.
This implementation supports OpenStack Swift and SoftLayer Object store

### DIFF
--- a/src/couch_att.erl
+++ b/src/couch_att.erl
@@ -438,17 +438,28 @@ digest_from_json(Props) ->
 
 
 to_json(Att, OutputData, DataToFollow, ShowEncoding) ->
-    [Name, Data, DiskLen, AttLen, Enc, Type, RevPos, Md5] = fetch(
+    [AttExternal,AttExternalSize, AttExternalMD5] = couch_att:fetch([att_external,att_external_size,att_external_md5],Att),
+    [Name, Data, DiskLenTmp, AttLen, Enc, Type, RevPos, Md5] = fetch(
         [name, data, disk_len, att_len, encoding, type, revpos, md5], Att
     ),
+    case AttExternal of
+        "external" ->
+            DiskLen = AttExternalSize,
+            DigestProp = case AttExternalMD5 of
+                <<>> -> [];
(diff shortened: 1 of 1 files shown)


src/couch_att.erl
```diff
@@ -438,17 +438,28 @@ digest_from_json(Props) ->
...
+            DiskLen = AttExternalSize,
+            DigestProp = case AttExternalMD5 of
+                <<>> -> [];
+                Digest -> [{<<"digest">>, <<"md5-", Digest/binary>>}]
+            end;
+        _ ->
+            DiskLen = DiskLenTmp,
+            DigestProp = case base64:encode(Md5) of
+                <<>> -> [];
+                Digest -> [{<<"digest">>, <<"md5-", Digest/binary>>}]
```
kxepal: Can we not repeat ourself in Digest generation?

src/couch_att.erl
```diff
@@ -438,17 +438,28 @@ digest_from_json(Props) ->
 
 
 to_json(Att, OutputData, DataToFollow, ShowEncoding) ->
-    [Name, Data, DiskLen, AttLen, Enc, Type, RevPos, Md5] = fetch(
+    [AttExternal,AttExternalSize, AttExternalMD5] = couch_att:fetch([att_external,att_external_size,att_external_md5],Att),
```
kxepal: Why do fetch the same attachment twice?

kxepal:
Please fix style: 80 chars line length, commas/spaces, etc.

gilv:
@kxepal @rnewson 
I provided another patch, implementing most of the comments.
Here is what is not implemented in this patch:

1.  Remark on the AttFun. Not yet implemented. Original remark:  “AttFun could be actually an attachment data or some fold function? Something went wrong if you have to do this.” 
2.  usage of  couch_util:to_hex
3.  still no unitests

gilv:
@kxepal Can you please review the recent code?

PR #121, opened 2015-10-14
Silence logs during apps startup / teardown in eunit tests.

nickva:
COUCHDB-2855


nickva:
Fixed via config file in https://github.com/apache/couchdb/commit/c6b9e74370210bbd14c8535e8457d85a26ddd945

PR #125, opened 2015-11-05
Run tests with only the couch_db_plugin_test couch_epi plugin

jaydoane:
Save current couch_epi application environment plugins in context during test setup, and restore during teardown.

COUCHDB-2868

### DIFF
--- a/test/couch_db_plugin_tests.erl
+++ b/test/couch_db_plugin_tests.erl
@@ -34,8 +34,6 @@
 -include_lib("couch/include/couch_eunit.hrl").
 -include_lib("couch/include/couch_db.hrl").
 
--record(ctx, {pid, handle}).
-
 %% couch_epi_plugin behaviour
 
 app() -> test_app.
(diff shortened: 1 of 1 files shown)


jaydoane:
@iilyak, does this look reasonable to you?

test/couch_db_plugin_tests.erl
```diff
@@ -48,24 +48,18 @@ notify(_, _, _) -> ok.
...
+    ok = application:set_env(couch_epi, plugins, [?MODULE]),
+    ok = application:start(couch_epi),
+    SavedPlugins.
 
 setup() ->
     error_logger:tty(false),
-    start_epi(),
-    #ctx{handle = couch_epi:get_handle(couch_db)}.
+    SavedPlugins = start_epi(),
+    #ctx{handle = couch_epi:get_handle(couch_db), saved_plugins=SavedPlugins}.
```
iilyak: Please add space before and after `=` sign.

test/couch_db_plugin_tests.erl
```diff
@@ -48,24 +48,18 @@ notify(_, _, _) -> ok.
...
 
-teardown(#ctx{}) ->
-    Plugins = application:get_env(couch_epi, plugins, []),
-    application:set_env(couch_epi, plugins, Plugins -- [?MODULE]),
+teardown(#ctx{saved_plugins=SavedPlugins}) ->
```
iilyak: spaces

iilyak:
Maybe we should stop and unload couch_epi in teardown instead. Since we load and start it in setup. In this case there will be no need to remember current list of configured plugins.

kxepal:
I think it worth to add some test that cause failure for that case before fix it. Because, we didn't have such problem now and we eventually may break this again without any notice.

jaydoane:
@iilyak, I agree with your last comment, as it seems a bit strange to save the plugin state while tests are running.

@kxepal, the reason check_is_admin_no_match was failing downstream was because -- in addition to loading the couch_db_pluging_tests plugin that was meant to handle that case -- all the couch_epi plugins specified in Cloudant's couch_epi.config were being loaded, and one of them was inadvertently handling (and breaking) the test case. I think any time a unit test pulls in configuration data that it doesn't completely specify, we run the risk of it breaking something downstream. I could certainly create another plugin that simulates what was happening, but I don't think it will really help prevent this sort of thing in the future. By loading *only* the plugins we want to test, we force test behavior downstream to be the same. Does that make sense?

jaydoane:
The latest changes clean the setup/teardown code considerably, in addition to speeding up the test suite.

@kxepal, may I suggest that if we want an example test showing epi plugin collisions/interference, that we create a separate issue for that? I'm happy to tackle that if we can get this resolved...

iilyak:
+1

kxepal:
@jaydoane ok with separate issue. +1

PR #126, opened 2015-11-06
Remove lager and goldrush from default list of apps started by test_util:start_couch/2

jaydoane:
Also, create a common list of default apps that is started and stopped consistently

COUCHDB-2872

### DIFF
--- a/src/test_util.erl
+++ b/src/test_util.erl
@@ -34,6 +34,9 @@
 
 -record(test_context, {mocked = [], started = [], module}).
 
+-define(DEFAULT_APPS,
+        [inets, ibrowse, ssl, config, couch_epi, couch_event, couch]).
+
 srcdir() ->
     code:priv_dir(couch) ++ "/../../".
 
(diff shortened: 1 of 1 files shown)


kxepal:
Funny that we still start goldrush and lager for couch_log_lager_backend implicitly somewhere else.
@iilyak @eiri do you remember if this will not bite us somewhere else in future?

iilyak:
@kxepal: We have a very bad cyclic dependency between [couch_log](https://github.com/apache/couchdb-couch-log/blob/master/src/couch_log.erl#L47) and [couch_stats](https://github.com/apache/couchdb-couch-stats/blob/master/src/couch_stats.erl#L118). The problem could happen if you would make a typo in metric name and run a test suite. We need to test for that before we merge this (either manually or by implementing test case).

jaydoane:
@iilyak, I'm sorry for being so dense, but can you go into a little more detail how this patch might trigger the cyclic dependency you mentioned above?

iilyak:
@jaydoane: The cyclic dependency could be triggered with following:

1. make a syntactic error in `couch-log/priv/stats_descriptions.cfg` 
2. `couch_stats_aggregator` would try to reload metrics. The reload logic skips all metrics defined in a file we cannot read (using `file:consult`)
3. Any attempt to log anything using `couch_log:<level>` would cause [`couch_log:notice("unknown metric: ~p", [Name])`](https://github.com/apache/couchdb-couch-stats/blob/master/src/couch_stats.erl#L118)
4. `couch_log:notice` in it's turn calls [`couch_stats:increment()`](https://github.com/apache/couchdb-couch-log/blob/master/src/couch_log.erl#L47)
5. we would have an infinite loop which will bring the node down.

However you a right the problem I mentioned would exhibit itself regardless of this particular change.

iilyak:
> However you a right the problem I mentioned would exhibit itself regardless of this particular change

Are there any other obstacles?

jaydoane:
I have rebased. If there are no other objections, can we get a +1?

tonysun83:
I ran some make checks with your branch and the tests I ran pass. +1 for me.

nickva:
Also tested.  Used this patch to pick up pr changes:

```
-{couch,            "couch",            "f33f9fa1281fd6c50a86cb9853bd2d9ec2dd007c"},
+%{couch,            "couch",            "f33f9fa1281fd6c50a86cb9853bd2d9ec2dd007c"},
+{couch,            {url, "https://github.com/cloudant/couchdb-couch.git"},
+ "010704ad9ddfd8bbd28c91e6c87484ef25afa296"},
```

Ran on master branch of couchdb, Mac OS X, Erlang 18. All tests pass and output looked good.

kxepal:
+1

PR #128, opened 2015-11-20
Explicitly authorize test requests

jaydoane:
Create admin user and use its credentials for requests

Use http requests to create and delete test dbs to avoid potential db
name munging issues downstream

COUCHDB-2897

### DIFF
--- a/test/couchdb_mrview_cors_tests.erl
+++ b/test/couchdb_mrview_cors_tests.erl
@@ -24,8 +24,14 @@
     ]}}
 ]}).
 
+-define(USER, "mrview_cors_test_admin").
+-define(PASS, "pass").
+-define(AUTH, {basic_auth, {?USER, ?PASS}}).
+
+
 start() ->
     Ctx = test_util:start_couch([chttpd]),
+    ok = config:set("admins", ?USER, ?PASS, _Persist=false),
     ok = config:set("httpd", "enable_cors", "true", false),
     ok = config:set("vhosts", "example.com", "/", false),
     Ctx.
@@ -37,10 +43,13 @@ setup(PortType) ->
     config:set("cors", "credentials", "false", false),
     config:set("cors", "origins", "http://example.com", false),
 
-    Addr = config:get("httpd", "bind_address", "127.0.0.1"),
-    Host = "http://" ++ Addr ++ ":" ++ port(PortType),
+    Host = host_url(PortType),
     upload_ddoc(Host, ?b2l(DbName)),
     {Host, ?b2l(DbName)}.
+
+teardown(Ctx) ->
+    ok = config:delete("admins", ?USER, _Persist=false),
+    test_util:stop_couch(Ctx).
 
 teardown(PortType, {_Host, DbName}) ->
     delete_db(PortType, ?l2b(DbName)),
@@ -51,7 +60,7 @@ cors_test_() ->
         "CORS for mrview",
         {
             setup,
-            fun start/0, fun test_util:stop_couch/1,
+            fun start/0, fun teardown/1,
             [show_tests()]
         }
     }.
@@ -75,7 +84,7 @@ should_make_shows_request(_, {Host, DbName}) ->
     ?_test(begin
          ReqUrl = Host ++ "/" ++ DbName ++ "/_design/foo/_show/bar",
          Headers = [{"Origin", "http://example.com"},
-                    {"Access-Control-Request-Method", "GET"}],
+                    {"Access-Control-Request-Method", "GET"}, ?AUTH],
          {ok, _, Resp, Body} = test_request:get(ReqUrl, Headers),
          Origin = proplists:get_value("Access-Control-Allow-Origin", Resp),
          ?assertEqual("http://example.com", Origin),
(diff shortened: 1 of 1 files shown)


test/couchdb_mrview_cors_tests.erl
```diff
@@ -86,14 +95,25 @@ create_db(backdoor, DbName) ->
...
-    ok = fabric:create_db(DbName, [?ADMIN_CTX]).
+    {ok, Status, _, _} = test_request:put(db_url(DbName), [?AUTH], ""),
+    true = lists:member(Status, [201, 202]),
+    ok.
 
 delete_db(backdoor, DbName) ->
     couch_server:delete(DbName, [?ADMIN_CTX]);
 delete_db(clustered, DbName) ->
-    ok = fabric:delete_db(DbName, [?ADMIN_CTX]).
+    {ok, 200, _, _} = test_request:delete(db_url(DbName), [?AUTH]),
```
kxepal: DELETE /db may eventually return HTTP 202 response.

kxepal: btw, is there reason to switch from Erlang to HTTP API for this?

kxepal:
LGFM, except on moment, but you seems start to repeat yourself. We have quite a lot of tests where Admin Party implicitly assumed. 

May be it worth to make a single change for test_util to setup cluster without AP and fix all at once?

test/couchdb_mrview_cors_tests.erl
```diff
@@ -86,14 +95,25 @@ create_db(backdoor, DbName) ->
...
+    ok.
 
 
+host_url(PortType) ->
+    Addr = config:get("httpd", "bind_address", "127.0.0.1"),
```
kxepal: Better to make a switch for httpd/chttpd here as well since you may eventually try to test backdoor interface using chttpd port. Someday.

test/couchdb_mrview_cors_tests.erl
jaydoane: In the spirit of YAGNI, I would be inclined to assume 200 for successful DELETE response until otherwise necessary. Also, the docs don't mention 202:
http://docs.couchdb.org/en/latest/api/database/common.html?highlight=delete#delete--db

jaydoane: The reason for the switch from fabric db creation to the HTTP API is because in some downstream cases (specifically to support multiple tenants in one cluster), the DB name gets munged in a way to indicate to which user it belongs. This happens transparently through the HTTP API, but fabric knows nothing about it.

test/couchdb_mrview_cors_tests.erl
jaydoane: Again, the principle of YAGNI was what prevented me from making that change for the eventual possibility. However, I'm fine to add it :)

test/couchdb_mrview_cors_tests.erl
kxepal: Documentation is not updated for 2.0 actually, so YAGNI source is in erlang sources for now (:
Ok with fabric to HTTP replacement, though it strange that we do this with Erlang, not JS. Need to revisit this later.

test/couchdb_mrview_cors_tests.erl
kxepal: Thanks! (:

jaydoane:
@kxepal, I hear what you are saying about the Admin Party being assumed for many tests. However, at this point I have not taken a full inventory, and am only focusing on the suites which are failing downstream for various reasons. So far, the ones I've been able to fix by adding explicit authentication are just this one and https://github.com/apache/couchdb-chttpd/pull/91. I'm guessing that the other implicit Admin Party tests may not be doing things (like creating/deleting DBs) which would cause breakage downstream, where an Admin Party is less permissive than in the CouchDB case.

So, in short, I don't think there will be many more PRs such as this. The remaining tests that are failing downstream have much bigger problems, and are likely "unfixable" based on what they are trying to test. I think we'll just have to disable them in downstream builds and move on :(

jaydoane:
Also, I'm wondering what's going on with travis. It's been over 12 hours that the build has been pending...

kxepal:
@jaydoane ): ok
Travis pending is seems because of ASF tasks queue overflow again. This can last for a long time, suddenly.

test/couchdb_mrview_cors_tests.erl
jaydoane: Actually, I think the only way the host_url depends on PortType (which can be either 'clustered' or 'backdoor') is expressed via the port/1 function. I'm not sure how you would improve or generalize this.

Or did you mean db_url, which is only valid for clustered PortType? That could be generalized by adding a PortType parameter, but I didn't bother since it's only being used in the clustered cases.

test/couchdb_mrview_cors_tests.erl
jaydoane: Ah, good to know about Docs - thanks. I'll go ahead and make the change, as it's pretty minor and will indeed make the test more robust.

test/couchdb_mrview_cors_tests.erl
kxepal: Basically, I think:
```
db_url(DbName) when is_binary(DbName) -> db_url(binary_to_list(DbName));
db_url(DbName) when is_list(DbName) -> host_url(cluster) ++ "/" ++ DbName.
host_url(Interface) -> "http://" ++ bind_address(Interface) ++ ":" ++ port(Interface).
bind_address(cluster) -> config:get("chttpd", "bind_address", "127.0.0.1");
bind_address(backdoor) -> config:get("httpd", "bind_address", "127.0.0.1").
port(cluster) -> integer_to_list(mochiweb_socket_server:get(chttpd, port));
port(backdoor) -> integer_to_list(mochiweb_socket_server:get(couch_httpd, port)).
```

jaydoane: Ah, this makes perfect sense now; thanks!

jaydoane:
@kxepal, I believe the changes we discussed have been included in latest push.

kxepal:
+1

PR #129, opened 2015-11-22
2898 improve global changes tests

jaydoane:


kxepal:
Don't you actually hide an issue with e7b3ba4 commit? Slash character is used by global_changes for specific logic owner.

jaydoane:
Yes, it's possible that there's a bug downstream that's triggered by the "/" in the db name. It works correctly for one of the downstream builds, but fails in the other. @kxepal, if you prefer, I can omit that commit.

kxepal:
I think you should omit that commit in order to not forget about the bug and fix it in downstream (: There is no better reminder as build failure.

jaydoane:
@kxepal, I've removed the offending commit.

kxepal:
+1

PR #131, opened 2015-11-27
Always send application/json response type

rnewson:
Modern browsers (I tested Firefox, Safari, Chrome) will render an
application/json response body as text in a fixed width font. This
ease of display motivated the text/plain vs application/json
negotiation.

It is now obsolete. Let's always send application/json content type.

COUCHDB-2905

### DIFF
--- a/src/couch_httpd.erl
+++ b/src/couch_httpd.erl
@@ -970,25 +970,11 @@ send_chunked_error(Resp, Error) ->
 send_redirect(Req, Path) ->
      send_response(Req, 301, [{"Location", absolute_uri(Req, Path)}], <<>>).
 
-negotiate_content_type(Req) ->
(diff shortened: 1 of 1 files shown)


kxepal:
> /home/kxepal/projects/asf/couchdb/couch/src/couch_httpd.erl:973: Warning: variable 'Req' is unused

rnewson:
fixed.

kxepal:
Tested with IE. IE 8 as expected asks to download some file while IE Edge works as expected rendering the response. Opera also handles this right. So technically, we drop <=IE8 support with easy, may be IE9 also affected, but don't have it to test.

+1

rnewson:
thanks! there will be some tests to clean up (or delete) but this improves our api.

PR #132, opened 2015-11-28
add delegated auth

janl:


janl:
This is the 2.x version of https://github.com/apache/couchdb/pull/366

Goes with:
- https://github.com/apache/couchdb-chttpd/pull/97
- https://github.com/apache/couchdb/pull/373

kxepal:
I think this is not needed: the main interface for 2.0 is chttpd. For backdoor interface I'm not sure you'll find a use case for this feature. Also, duplicated code makes it hard to maintain.

PR #135, opened 2016-01-28
Add couch_file:open_files/0

iilyak:
Adding debugging function to list all opened couch_file files.
This function is for debugging purposes only. It iterates through all
open ports and returns file descriptor and filename correspondent to
every port.

### DIFF
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -42,6 +42,8 @@
 -export([init/1, terminate/2, code_change/3]).
 -export([handle_call/3, handle_cast/2, handle_info/2]).
 
+%% helper functions
+-export([process_info/1]).
 
 %%----------------------------------------------------------------------
 %% Args:   Valid Options are [create] and [create,overwrite].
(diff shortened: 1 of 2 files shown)


src/couch_file.erl
```diff
@@ -299,6 +301,8 @@ init({Filepath, Options, ReturnPid, Ref}) ->
         filelib:ensure_dir(Filepath),
         case file:open(Filepath, OpenOptions) of
         {ok, Fd} ->
+            %% Save Fd in process dictionary for debugging purposes
+            put(couch_file_fd, {Fd, Filepath}),
```
kxepal: Why not to use `pid2name/1` instead?

kxepal:
Good call to provide debug helper utilities! 
A bit long shot, but I think better to start with separate module (like `couch_debug`). This function won't be the only and we can keep those isolated from modules public api (since comment about debugging purposes won't be available on autocompletion hints). And it could make us free on utilities content. What do you think?

src/couch_file.erl
iilyak: pid2name doesn't work for raw file descriptors.

kxepal: Hm...true, I miss that moment.

tsutsu:
I wouldn't expect a function called `couch_file:process_info(Pid)` to return a single property tuple, but rather a processed and filtered version of `erlang:process_info(Pid)`—a list of tuples.

What you probably want is more like `erlang:process_info(Pid, Key)`—a key parameter to make it clear you're pulling out a particular tuple, e.g. `couch_file:process_info(Pid, fd)`. The function could just be defined for the `fd` case and nothing else.

Other than that, looks good.

iilyak:
@tsutsu: Thank you.

iilyak:
@tsutsu: Your suggestion was implemented.

tsutsu:
+1

PR #136, opened 2016-01-29
Add should_close_old_fd_after_view_compaction test

iilyak:
Test case for PR https://github.com/apache/couchdb-couch-mrview/pull/39


eiri:
Works as expected. +1 after apache/couchdb-couch-mrview#39 lands.

kxepal:
Ah, I think I even now that problem as disaster causer. Nice found! +1

PR #138, opened 2016-02-09
2945 remove couch http cors

iilyak:
Remove second implementation of CORS (couch_http_cors) in favor of chttpd_cors. This refactoring is done in order to make sure we always call chttpd_cors:headers for backdoor interface as well as chttpd interface. 

This PR depends on https://github.com/apache/couchdb-chttpd/pull/101


kxepal:
I'm all +1 for this, except that I'm not sure if this is correct design-wise solution. Because we actually bring bits from cluster httpd interface to the backdoor one. And since chttpd uses couch_httpd internally, this makes circular dependency between them. 

What I think would be correct is to move common httpd bits into own package. We already have couch-httpd repository as empty and it name sounds to be good place for such kind of things. And then we can have
```
     couch-httpd
     ^         ^
     |         |
  chttpd couch_httpd
```

model instead of 
```
chttpd <--> couch_httpd
```

In mid/long term couch-httpd app can help to isolate mochiweb from rest CouchDB and simplify migration to some else web framework: webmachine, cowboy, whatever it'll be.

What do you think about this?

iilyak:
@kxepal: https://github.com/apache/couchdb-couch-httpd is not empty. I'm not brave enough to pull off 3-ways refactoring in a reasonable time. Are there any plans to keep the current content of `couchdb-couch-httpd`?

kxepal:
@iilyak Oh, I forgot about those. That content was a part of contributed RCouch fork, where couch_http* modules were moved into own app. 

Currently it doesn't get used anyhow. I don't remember if there are any valuable features to backport (except refactoring), but since it's isolated into own branch, I think we can use use master branch for our needs while rcouch bits remains untouched till we figure out what to do with them.

@rnewson @davisp @chewbranca @eiri - what's your opinion on this?

eiri:
My understanding is that this refactoring addressing the problem we have with CORS for some of end-points (_sessions ?) that hasn't been fixed by previous CORS' PR?

On a big scale of things I agree that having common couch-httpd lib is a great idea, but if this change is not a cleanup, but a fix, I think we can do that after getting it in, not instead.

davisp:
If this fixes a specific issue then we should call it out. The commit message is vague enough it sounds like just a style/debt thing. I think we're all in agreement that we want to dedupe everything and this is only a very small part of that. If this is just a small part of the dedupe then I'd delay this until we do the whole dedupe. However if its a fix for a problem them I'm fine pulling it in so long as we make sure and explain the specific problem we're solving so that future us are more prepare when we get to the eventual dedupe work.

chewbranca:
@kxepal @iilyak @eiri @davisp in general I think the approach of deleting `couch_httpd_cors` in favor of `chttpd_cors` is correct and is fine to do here. The `chttpd_cors` implementation was based off of `couch_httpd_cors` with a number of changes cleaning things up and improving the implementation. We talked about whether to remove `couch_httpd_cors` in the past but the `chttpd_cors` implementation did not support the same multi-host config format that already existed, a problem @iilyak has rectified in this set of PRs, so in general I'm +1 to this change.

For this PR, does it make sense to delete the CORS implementation but leave in the CORS tests?

As for introducing a third `*_httpd` library layer, I think we would be better off spending the time consolidating down to a single HTTP stack and removing the current dichotomy as quickly as we can.

iilyak:
> @chewbranca: For this PR, does it make sense to delete the CORS implementation but leave in the CORS tests?

Traditionally we put integration tests related to other repositories into `couchdb-couch` application [see chttpd_endpoints_tests.erl](https://github.com/apache/couchdb-couch/blob/master/test/chttpd_endpoints_tests.erl). So the old CORS test suite is just lifted to become an integration suite.

kxepal:
I wish to break this tradition, but don't have enough strength and will to rework whole the tests once again. For now it doesn't matter where to place integration tests since for each app we build full CouchDB to run the tests. So integration tests could be places wherever, but I would agree that it's quite strange to see chttpd_* tests in couch app and vice versa.

chewbranca:
+1

PR #139, opened 2016-02-11
Propagate db delete event to changes callback

iilyak:
Not propagating the `delete` event to the changes callback causes
db to stay open when it is deleted in the presence of `continuous`
requests to `_changes` feed. This in its turn causes `couch_file` to stay
open until the connection is closed by the client.

### DIFF
--- a/src/couch_changes.erl
+++ b/src/couch_changes.erl
@@ -170,7 +170,9 @@ handle_changes(Args1, Req, Db0, Type) ->
 handle_db_event(_DbName, updated, Parent) ->
     Parent ! updated,
     {ok, Parent};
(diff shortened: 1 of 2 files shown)


davisp:
Ahhh, took me two readings of your commit message before I put the whole thing together. I kept trying to figure out how we introduced the bug during the merge and got confused by the work on view changes. Rereading your commit message it appears to not be a new bug, just one we hadn't noticed yet.

+1

davisp:
I looked at that test failure and it appears unrelated, but we might also want to verify that as well.

kxepal:
+1

PR #140, opened 2016-02-15
Fix test_util:with_process_restart

iilyak:
- The order of arguments is incorrect.
- Return the result of stop_sync from with_process_restart/3


mikewallace1979:
There's a typo in the commit title (`precess`) - other than that, +1

kxepal:
+1

PR #141, opened 2016-02-16
Refactor rename on delete

iilyak:
The goals of this PR.
- move `rename_on_delete` feature into `couch_file`
- avoid calling `config` from `couch_file`

### DIFF
--- a/src/couch_db_updater.erl
+++ b/src/couch_db_updater.erl
@@ -43,9 +43,10 @@ init({DbName, Filepath, Fd, Options}) ->
         ok = couch_file:write_header(Fd, Header),
         % delete any old compaction files that might be hanging around
         RootDir = config:get("couchdb", "database_dir", "."),
-        couch_server:delete_file(RootDir, Filepath ++ ".compact"),
-        couch_server:delete_file(RootDir, Filepath ++ ".compact.data"),
-        couch_server:delete_file(RootDir, Filepath ++ ".compact.meta");
+        DelOpts = [compaction],
+        couch_server:delete_file(RootDir, Filepath ++ ".compact", DelOpts),
+        couch_server:delete_file(RootDir, Filepath ++ ".compact.data",DelOpts),
+        couch_server:delete_file(RootDir, Filepath ++ ".compact.meta",DelOpts);
     false ->
         case couch_file:read_header(Fd) of
         {ok, Header} ->
@@ -105,7 +106,8 @@ handle_call(cancel_compact, _From, #db{compactor_pid = Pid} = Db) ->
     unlink(Pid),
     exit(Pid, kill),
     RootDir = config:get("couchdb", "database_dir", "."),
-    {ok, _} = couch_server:delete_file(RootDir, Db#db.filepath ++ ".compact"),
+    {ok, _} = couch_server:delete_file(RootDir, Db#db.filepath ++ ".compact",
+        [compaction]),
     Db2 = Db#db{compactor_pid = nil},
     ok = gen_server:call(couch_server, {db_updated, Db2}, infinity),
     {reply, ok, Db2};
@@ -246,11 +248,12 @@ handle_cast({compact_done, CompactFilepath}, #db{filepath=Filepath}=Db) ->
                         [Filepath, CompactFilepath]),
         ok = file:rename(CompactFilepath, Filepath ++ ".compact"),
         RootDir = config:get("couchdb", "database_dir", "."),
-        couch_server:delete_file(RootDir, Filepath),
+        couch_server:delete_file(RootDir, Filepath, [compaction]),
         ok = file:rename(Filepath ++ ".compact", Filepath),
         % Delete the old meta compaction file after promoting
         % the compaction file.
-        couch_server:delete_file(RootDir, Filepath ++ ".compact.meta"),
+        couch_server:delete_file(RootDir, Filepath ++ ".compact.meta",
+            [compaction]),
         close_db(Db),
         NewDb3 = refresh_validate_doc_funs(NewDb2),
         ok = gen_server:call(couch_server, {db_updated, NewDb3}, infinity),
--- a/src/couch_db_updater.erl
+++ b/src/couch_db_updater.erl
@@ -43,9 +43,9 @@ init({DbName, Filepath, Fd, Options}) ->
         ok = couch_file:write_header(Fd, Header),
         % delete any old compaction files that might be hanging around
         RootDir = config:get("couchdb", "database_dir", "."),
-        couch_file:delete(RootDir, Filepath ++ ".compact"),
-        couch_file:delete(RootDir, Filepath ++ ".compact.data"),
-        couch_file:delete(RootDir, Filepath ++ ".compact.meta");
+        couch_server:delete_file(RootDir, Filepath ++ ".compact"),
+        couch_server:delete_file(RootDir, Filepath ++ ".compact.data"),
+        couch_server:delete_file(RootDir, Filepath ++ ".compact.meta");
     false ->
         case couch_file:read_header(Fd) of
         {ok, Header} ->
@@ -105,7 +105,7 @@ handle_call(cancel_compact, _From, #db{compactor_pid = Pid} = Db) ->
     unlink(Pid),
     exit(Pid, kill),
     RootDir = config:get("couchdb", "database_dir", "."),
-    ok = couch_file:delete(RootDir, Db#db.filepath ++ ".compact"),
+    {ok, _} = couch_server:delete_file(RootDir, Db#db.filepath ++ ".compact"),
     Db2 = Db#db{compactor_pid = nil},
     ok = gen_server:call(couch_server, {db_updated, Db2}, infinity),
     {reply, ok, Db2};
@@ -246,11 +246,11 @@ handle_cast({compact_done, CompactFilepath}, #db{filepath=Filepath}=Db) ->
                         [Filepath, CompactFilepath]),
         ok = file:rename(CompactFilepath, Filepath ++ ".compact"),
         RootDir = config:get("couchdb", "database_dir", "."),
-        couch_file:delete(RootDir, Filepath),
+        couch_server:delete_file(RootDir, Filepath),
         ok = file:rename(Filepath ++ ".compact", Filepath),
         % Delete the old meta compaction file after promoting
         % the compaction file.
-        couch_file:delete(RootDir, Filepath ++ ".compact.meta"),
+        couch_server:delete_file(RootDir, Filepath ++ ".compact.meta"),
         close_db(Db),
         NewDb3 = refresh_validate_doc_funs(NewDb2),
         ok = gen_server:call(couch_server, {db_updated, NewDb3}, infinity),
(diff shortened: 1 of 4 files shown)


src/couch_file.erl
```diff
@@ -216,24 +216,42 @@ close(Fd) ->
     gen_server:call(Fd, close, infinity).
 
 
-delete(RootDir, Filepath) ->
-    delete(RootDir, Filepath, true).
-
-
-delete(RootDir, Filepath, Async) ->
+delete(RootDir, Filepath, _Async, true) ->
```
eiri: @iilyak I think we'd better just to make third attribute here into options' list and pass both async and rename_on_delete flags in it, instead of growing arity. or am I missing something?

iilyak: Yes we can do that.

eiri:
@iilyak Ok, here we go. I've got a bit carried away trying to make `delete_file` to distinguish delete from compaction from context, but at the end it got too convoluted, so I threw it away and went with just adding a new delete option 'compaction'.  This is how it works now when 'rename_on_delete' enabled:

1. On database delete db file got renamed the same way as it was before - with _.YMD.HMS.deleted_ suffix and kept in a shard directory.
1. The deleted database's ddocs directories got renamed with the same suffix and stay in according shards dirs. The view files inside not renamed. PR for this : apache/couchdb-couch-index#15
1. On database compaction pre-compacted db file got renamed to `$shard:$dbname:YMD.HMS.$uuid.couch` and moved into `.delete` directory, but not deleted. This is partly done to avoid mixes up on db restoration, so old compactions will not be confused with honest deleted file.
1. Compaction's meta file treated the same and moved into `.delete` as well, except its extension kept _.meta_
1. On views compaction view files got renamed to `$shard:$ddoc_dir:$filename.YMD.HMS.$uuid.view` and moved into `.delete` directory of views root.
1. As confusing as it is, there are two separate approaches to views cleanups. One is in fabric, but not marshalled through rexi. It's just plain brutally unlinks all tombstoned view files on local to it node. As far as I understand it's a desperate measure to disk cleanup. I left it as it is.
1. Another one in `couch_mrview_cleanup.erl`. It seems it intended to be used with API end-point `/db/_view_cleanup`, but I'm not sure if it's actually on clustered interface. Anyway, I left it to act as delete, not compaction, so it'll be renaming tombstoned view files right in place, in shards dirs, the same way as database files renamed on delete.

I tried to keep commits in logical order, I know you prefer to review them that way. The thing got a bit larger than your original PR, so let me know what do you think.

src/couch_db_updater.erl
```diff
@@ -43,9 +43,10 @@ init({DbName, Filepath, Fd, Options}) ->
         ok = couch_file:write_header(Fd, Header),
         % delete any old compaction files that might be hanging around
         RootDir = config:get("couchdb", "database_dir", "."),
-        couch_file:delete(RootDir, Filepath ++ ".compact"),
-        couch_file:delete(RootDir, Filepath ++ ".compact.data"),
-        couch_file:delete(RootDir, Filepath ++ ".compact.meta");
+        DelOpts = [compaction],
+        couch_server:delete_file(RootDir, Filepath ++ ".compact", DelOpts),
+        couch_server:delete_file(RootDir, Filepath ++ ".compact.data",DelOpts),
```
iilyak: Something wrong with indentation

eiri: this one is a char longer than 80. our common approach is to squeeze spaces in such case, as it seems from codebase, rather than adding a line.

iilyak: I am not sure I like the approach. Let's see what others say.

iilyak: We code use the same approach as [here](https://github.com/apache/couchdb-couch/blob/master/src/couch_server.erl#L457:L459). We could also extract this common code into helper function.
While doing so I would also remove [this line](https://github.com/apache/couchdb-couch/blob/master/src/couch_server.erl#L460). Since the very same file is removed couple of lines above.

iilyak:
Thank you @eiri. I like the direction it is going to.

src/couch_db_updater.erl
eiri: yes, good idea

eiri:
@iilyak I've addressed all the comments, but rebase killed all the github discussions that was happening on the commits and not on the main PR diff, really sorry about that.

Is there anything else you don't like in the proposed changes?

iilyak:
@eiri: There are no tests for none clustered dbs.

iilyak:
@eiri: I tried to understand what the tokenizer do and it was quite hard. I had a doubt that it does work as intended in case of none clustered database. Therefore I added the tests and discover that it indeed doesn't pass. It might be possible that my test cases are wrong though. I'm ok with existent implementation as long as you would add tests make them pass and factor out the tokenizer into a function. I do agree with you that it is quite hard to make it easier to understand because it is special case upon another special case.

iilyak:
> @eiri https://github.com/cloudant/couchdb-couch/pull/11#issuecomment-199885171
To be honest I don't find your version much clearer on what it is doing or more idiomatic than mine foldl. It might be jiffy faster, but we are talking about traversing 5-6 elements list...

The point was not about performance but to improve readability by eliminating magic numbers.

> I guess we can refactor that tokenizer into its own function, though it's not like it's going to be used outside of deleted_filename. So what exactly is the gain of a hair splitting here?

Extracting black magic into a function so it is easier to find and fix later.

iilyak:
Don't get me wrong I like this work a lot. Let's just add tests and be done with it.

iilyak:
@eiri: We might need to replace `:` with something else to support the feature on windows.

eiri:
hm. I thought windows doesn't care about `:`, just macos 9 does? don't have windows handy to check out...

eiri:
[bummer](http://superuser.com/questions/358855/what-characters-are-safe-in-cross-platform-file-names-for-linux-windows-and-os) Well, how about number sign `#`?

iilyak:
@eiri: `#` could work. Unfortunately we don't validate user so if username contains `#` it would be tricky. We could do `couch_util:url_encode("/foo/bar")`. [Unreserved characters](https://tools.ietf.org/html/rfc3986#section-2.3) look portable to me if used as part of a file name. I.e. we would have `.delete/<url_encode(path)>.<timestamp>.<filetype>.deleted`. We don't need `.deleted` suffix and can get rid of it probably.

eiri:
@iilyak username goes into db name and we do validate db names. try for kicks `http PUT :5984/bird%23bird%2Fsparrow`. Unencoded `#` will be treated as URL anchor, `http PUT :5984/bird#bird%2Fsparrow` will just create db `bird`, so we safe on that matter.

There are no `.deleted` suffix on files moved into `.delete` directory, i.e. on files removed on compaction. `.deleted` added only on 'really' deleted files that stay in a shards directory and I think we need to keep it to simplify periodic cleanup, or backup, or whatever it is an alternative to "delete on a spot" strategy.

Anyway, I'll got with `#` as a separator. I'm finishing with less confusing, more extended tests here, going to push update in a couple of hours.

eiri:
I've redone `deleted_filename` with regexps to avoid magic filters. Granted it still not exactly obvious piece of code, but I hope it's more clear now what's going on in there.

I've added edge cases to the tests and made them more explicit on how file hierarchy organized for db and view files.

eiri:
and now without silly mistake. though that made me think if we want to be more accepting and do something like a fail-back on just substituting all `/` on `#` in the files that don't match our regexp.

/cc @iilyak

src/couch_file.erl
```diff
@@ -218,24 +218,63 @@ close(Fd) ->
...
+    io_lib:format("~s.~s.~s~s", [RootName, Suffix, Ending, Extention]).
+
+deleted_filename_base(RootDir, FullPath) ->
+    Shard = "\\.?shards/([0-9a-f]{8}-[0-9a-f]{8})",
+    User = "\\.?(.+)",
+    ViewFile = "(?:mrview/(.+\\.view))",
+    DbFile = "(.+\\.couch.*)",
+    Pattern = "^~s(?:/~s)?(?:/~s)?(?:/(?:~s|~s))$",
+    RegExp =  io_lib:format(Pattern, [RootDir, Shard, User, ViewFile, DbFile]),
+    {ok, MP} = re:compile(RegExp),
```
iilyak: Compiling the regexp would improve performance of a regexp if it is done once and reused multiple times. Here it looks like we compile it every time.

eiri: right. where are you suggesting to keep this compiled regexp? `delete` is just a function, it's not on `couch_file` gen_server's loop

iilyak: The common way to deal with one off calls is to just use `re:run`.

eiri: I assume you know that just re:run does the very same thing - passes params to BIF that either compiles regex and runs match or detects MP and runs the match right away, so it is not about performance. And I doubt this particular part of code is hard to understand. So I should admit I don't really get what is your rational on rejecting this change now, but given it's a third implementation I'm doing I don't really care anymore.

Are you agree with just substituting file path with `#` for those files and be done with it?

iilyak: > so it is not about performance. 

It is not about performance. Sorry I was not clear enough. I just wanted to mention that the `re:compile` line is not needed in this case.

> So I should admit I don't really get what is your rational on rejecting this change now, but given it's a third implementation I'm doing I don't really care anymore.

I'm sorry for being such a jerk. I didn't meant to. 

> Are you agree with just substituting file path with # for those files and be done with it?

Either `#` or urlencode as @kxepal proposed. As a primary mechanism (not as a fallback). It would simplify implementation a lot.

eiri:
I went with `couch_util:url_encode/1` which makes longer names, but reduces magic factor and makes automatic un-deletion (if needed) a bit more straightforward. It is `couch_util` flavour and not `ibrowse_lib`'s to not to deal with possible `+` in the names, which later uses to encode space. I've kept leading slash in the moved names to avoid special dealing with a leading dot.

src/couch_file.erl
```diff
@@ -247,7 +247,7 @@ delete_file(FilePath) ->
 deleted_filename(_RootDir, Original, rename) ->
     deleted_filename(Original, "deleted");
 deleted_filename(RootDir, Original, _Strategy) ->
-    DelFileName = deleted_filename_base(RootDir, Original),
+    DelFileName = couch_util:url_encode(Original -- RootDir),
```
iilyak: I have a concern with using `--` here. I know that it is a common thing in erlang community to do something like this. But it so error prone. There are at least two problems:
1. number of slashes makes the difference
2. it brakes badly if two paths do not have common prefix
```
"/foo/bar/baz" -- "/src/bar".
"foo/baz"
```
Feel free to ignore this. Since this is my complain to the state of erlang's stdlib which is the mess.

iilyak:
I am +1 for this change. However since I also contributed to this work it would be nice if someone else would review it as well.

rnewson:
This is a lot to read for something apparently simple. It's also not a refactor as it changes behaviour, so please correct the title and any comments claiming its a refactor.

Will try to devote some time to this soon, but, really, it's this complicated?

rnewson:
I think this can and should be much simpler. I'm not sure where the reluctance to call `config:get` in `couch_file:delete` comes from but I don't find it objectionable and it will radically simplify this change.

I note, as it's not noted anywhere so far, that the intention of the patch is to allow the administrator to decide if compacted files are immediately deleted or just renamed to the `.delete` directory (for removal by an external process). Cloudant does this to avoid performance issues when deleting large files.

eiri:
@iilyak Will it be ok for me to close this PR and its companions and create a new one, with simplified approach? History here is somehow cluttered and PR's title is misleading on what we are doing here.

iilyak:
@eiri: Sure

PR #142, opened 2016-02-22
fix tarball build

janl:
Seems to have no effect on clang-based systems, but should fix gcc builds.

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -32,7 +32,12 @@ CouchJSName = case os:type() of
         "couchjs"
 end,
 CouchJSPath = filename:join(["priv", CouchJSName]),
-Version = string:strip(os:cmd("git describe --always"), right, $\n),
+Version = case os:getenv("COUCHDB_VERSION") of
+    false ->
+        string:strip(os:cmd("git describe --always"), right, $\n);
+    Version0 ->
+        Version0
+end,
 
 CouchConfig = case filelib:is_file(os:getenv("COUCHDB_CONFIG")) of
     true ->


janl:
Requires https://github.com/apache/couchdb/pull/385

rebar.config.script
```diff
@@ -32,7 +32,15 @@ CouchJSName = case os:type() of
         "couchjs"
 end,
 CouchJSPath = filename:join(["priv", CouchJSName]),
-Version = string:strip(os:cmd("git describe --always"), right, $\n),
+Version0 = string:strip(os:cmd("git describe --always"), right, $\n),
+% When we are not building from a git checkout, the above will fail.
+% If that is the case, we use the COUCHDB_VERSION variable that has
+% been set by the top level Makefile
+Version = case Version0 of
+    "fatal:" ++ _Msg ->
```
kxepal: If there are no git on the target system, this message will be different:
```
bash: git: command not found
```

Other shells may give different error messages. May be use another strategy here:
- If COUCHDB_VERSION is set - use it
- Else try call git describe
- If success (exit status 0) - use that value
- Otherwise set version as "unknown" and cry loudly

janl: good point!

kxepal: Hm...but how we'll transfer COUCHDB_VERSION value to the tarball in this case? I'd tried to solve that problem, but everytime it ended that I had to create some version file on make dist step and read the version from there. Better was some pregenerated rebar.conf.script, but it turns to be more complicated as I hoped.

janl: hm, how do I get the exit status from `os:cmd()`?

kxepal: I fear you have to open port for that. Turns to be complicated a bit.

janl: > Hm...but how we'll transfer COUCHDB_VERSION value to the tarball in this case?

It looks like I got fed up with trying to solve this programmatically, so I made this: https://github.com/apache/couchdb/blob/master/version.mk — e.g.: we have to maintain that file by hand. like we do with `acinclude.m4` in 1.x.x: https://github.com/apache/couchdb/blob/1.x.x/acinclude.m4.in#L19-L21

janl: > > hm, how do I get the exit status from os:cmd()?
> I fear you have to open port for that. Turns to be complicated a bit.

yeah, let’s ignore this for now.

kxepal: Ah, there is such file! Thanks, I totally miss it.

janl:
pushed new version

kxepal:
+1 it works!

kxepal:
Another issue found: we don't include our rebar into tarball so ./configure tries to fetch and build it. I'll make a patch for that.

janl:
cool, thanks!

janl:
merged, thanks

PR #144, opened 2016-02-25
Conditionally use fetch and ddoc_cache logic

chewbranca:
This fixes issues seen in https://github.com/apache/couchdb-couch-replicator/pull/26.

### DIFF
--- a/src/couch_changes.erl
+++ b/src/couch_changes.erl
@@ -231,11 +231,15 @@ configure_filter(FilterName, Style, Req, Db) ->
     FilterNameParts = string:tokens(FilterName, "/"),
     case [?l2b(couch_httpd:unquote(Part)) || Part <- FilterNameParts] of
         [DName, FName] ->
-            DesignId = <<"_design/", DName/binary>>,
-            {ok, DDoc} = ddoc_cache:open_doc(fabric:dbname(Db), DesignId),
+            {ok, DDoc} = open_ddoc(Db, <<"_design/", DName/binary>>),
             check_member_exists(DDoc, [<<"filters">>, FName]),
-            DIR = fabric_util:doc_id_and_rev(DDoc),
-            {fetch, Style, Req, DIR, FName};
+            case Db#db.id_tree of
+                undefined ->
+                    DIR = fabric_util:doc_id_and_rev(DDoc),
+                    {fetch, Style, Req, DIR, FName};
+                _ ->
+                    {custom, Style, Req, DDoc, FName}
+            end;
 
         [] ->
             {default, Style};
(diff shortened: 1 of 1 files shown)


chewbranca:
Companion PR out in https://github.com/apache/couchdb-chttpd/pull/103.

kxepal:
Given the yesterday conversation on IRC, this all happens because we have a problem with design. Actually, that we have backdoor interface as standalone entity which eventually receives cluster bits that it can't eat. The best solution will require a lot of work and global things refactoring => more time.

Since this hack helps to solve things in short term, I'm +1, but only if we won't let it live for too long. And since backdoor interface have to go in anyway, so shall it be.

chewbranca:
Agreed @kxepal. It would be good to hear from @eiri as to whether this fixes the issues he's scene in https://github.com/apache/couchdb-couch-replicator/pull/26.

src/couch_changes.erl
```diff
@@ -221,21 +226,25 @@ configure_filter("_view", Style, Req, Db) ->
...
     FilterNameParts = string:tokens(FilterName, "/"),
     case [?l2b(couch_httpd:unquote(Part)) || Part <- FilterNameParts] of
         [DName, FName] ->
-            DesignId = <<"_design/", DName/binary>>,
-            {ok, DDoc} = ddoc_cache:open_doc(fabric:dbname(Db), DesignId),
+            {ok, DDoc} = open_ddoc(Db, <<"_design/", DName/binary>>),
             check_member_exists(DDoc, [<<"filters">>, FName]),
-            DIR = fabric_util:doc_id_and_rev(DDoc),
-            {fetch, Style, Req, DIR, FName};
+            case UseFetch of
```
eiri: I believe we can use the same logic as in `open_ddoc/2` to determine if we want `fetch` or `custom`, i.e.
```
            case Db#db.id_tree of
                undefined ->
                    DIR = fabric_util:doc_id_and_rev(DDoc),
                    {fetch, Style, Req, DIR, FName};
                _ ->
                    {custom, Style, Req, DDoc, FName}
            end;
```
Granted, it's less explicit than using `UseFetch` tag, but consider that we want to get rid of the whole none-clustered business it worth of trade - we'll need to change less code and wouldn't need to touch couchdb-chttpd at all.

How do you think?

src/couch_changes.erl
```diff
@@ -339,15 +348,8 @@ check_docids(_) ->
...
-    {_, Ref} = spawn_monitor(fun() ->
-        exit(fabric:open_doc(mem3:dbname(DbName), DDocId, [ejson_body]))
-    end),
-    receive
-        {'DOWN', Ref, _, _, {ok, _}=Response} ->
-            Response;
-        {'DOWN', Ref, _, _, Response} ->
-            throw(Response)
-    end;
+    {ok, DDoc} = ddoc_cache:open_doc(mem3:dbname(DbName), DDocId),
```
eiri: You don't need to repeat output here, the assign will return `{ok, DDoc}` or throw on none-match

src/couch_changes.erl
chewbranca: Yeah sounds good to me. I wasn't a huge fan of adding `configure_filter/5`, so happy to switch it around.

src/couch_changes.erl
chewbranca: Sure, I usually like to be explicit about it, but switching to `{ok, _DDoc}` is fine with me.

chewbranca:
@kxepal @eiri alright I rewrote the patch using the `Db#db.id_tree == undefined` hack and rebased out the old commit.

kxepal:
Nice!

eiri:
+1 :shipit:

PR #147, opened 2016-02-27
Add counters for lifetime total queued and dequeued items

tsutsu:
This PR enables a work queue to keep a running total of all the work that's ever been added to it, queryable through the functions `total_in/1`, `total_out/1`, and `total_held/1`. Additionally, a `queue(Wq, Item, CountsAs)` function-clause has been added and exported, which enables a single queue-item to be counted as an arbitrary number of work-items from the perspective of the running total counters. This allows for the ability to track the total number of individual work-items processed, even when queuing and dequeuing are done in "chunks" of items, as is done in `couch_index_updater`.

COUCHDB-2955
BugzID: 14036

### DIFF
--- a/src/couch_work_queue.erl
+++ b/src/couch_work_queue.erl
@@ -17,7 +17,7 @@
 -include_lib("couch/include/couch_db.hrl").
 
 % public API
--export([new/1, queue/2, dequeue/1, dequeue/2, close/1, item_count/1, size/1]).
+-export([new/1, queue/2, queue/3, dequeue/1, dequeue/2, close/1, item_count/1, size/1, total_in/1, total_out/1, total_held/1]).
 
 % gen_server callbacks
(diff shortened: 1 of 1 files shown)


src/couch_work_queue.erl
```diff
@@ -17,7 +17,7 @@
 -include_lib("couch/include/couch_db.hrl").
 
 % public API
--export([new/1, queue/2, dequeue/1, dequeue/2, close/1, item_count/1, size/1]).
+-export([new/1, queue/2, queue/3, dequeue/1, dequeue/2, close/1, item_count/1, size/1, total_in/1, total_out/1, total_held/1]).
```
kxepal: May be wrap this to fit 80 (at least) line right margin?

kxepal:
Here need some tests to make sure that it counts right.

PR #133, opened 2016-01-20
Add chttpd_plugin:before_response/4 hook

iilyak:
This hook is useful in following casses:

  - to inject vendor specific headers
  - to inject vendor keys into json objects returned to client
  - to override return code

PR #150, opened 2016-03-10
2966 before response hook

iilyak:
Implement `before_response` and `before_serve_file` EPI hooks.

These hooks are useful in following cases:

 - to inject vendor specific headers
 - to inject vendor keys into json objects returned to client
 - to override return code

This is a replacement PR for #133

### DIFF
--- a/src/couch_httpd.erl
+++ b/src/couch_httpd.erl
@@ -659,10 +659,8 @@ log_response(Code, Body) ->
     end.
 
 start_response_length(#httpd{mochi_req=MochiReq}=Req, Code, Headers0, Length) ->
-    log_request(Req, Code),
-    couch_stats:increment_counter([couchdb, httpd_status_codes, Code]),
     Headers1 = basic_headers(Req, Headers0),
-    Resp = MochiReq:start_response_length({Code, Headers1, Length}),
+    Resp = handle_response(Req, Code, Headers1, Length, start_response_length),
     case MochiReq:get(method) of
     'HEAD' -> throw({http_head_abort, Resp});
     _ -> ok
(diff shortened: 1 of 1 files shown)


src/couch_httpd.erl
```diff
@@ -1120,6 +1110,37 @@ validate_bind_address(Address) ->
...
+
+before_response(Req0, Code0, Headers0, {json, JsonObj}) ->
+    {ok, {Req1, Code1, Headers1, Body1}} =
+        chttpd_plugin:before_response(Req0, Code0, Headers0, JsonObj),
+    Body2 = [start_jsonp(), ?JSON_ENCODE(Body1), end_jsonp(), $\n],
+    {ok, {Req1, Code1, Headers1, Body2}};
+before_response(Req0, Code0, Headers0, Args0) ->
+    chttpd_plugin:before_response(Req0, Code0, Headers0, Args0).
+
+respond_(#httpd{mochi_req = MochiReq}, Code, Headers, _Args, start_response) ->
```
kxepal: Does it worth to isolate MochiReq here and now while we still have a lot of places where we use it directly? I just quite don't like such kind of generic functions and it seems that this doesn't gives any positive gain now, but one more function in call chain (and traceback).

kxepal:
+1, but I think we can live without `respond_`.

src/couch_httpd.erl
kxepal: Ok, after thinking about the ways how to make the same changes without it, I ended with callbacks. Now sure which one is better here. So ok.

src/couch_httpd.erl
```diff
@@ -476,14 +476,18 @@ accepted_encodings(#httpd{mochi_req=MochiReq}) ->
...
 
-serve_file(#httpd{mochi_req=MochiReq}=Req, RelativePath, DocumentRoot,
-           ExtraHeaders) ->
-    log_request(Req, 200),
-    ResponseHeaders = server_header()
-        ++ couch_httpd_auth:cookie_auth_header(Req, [])
-        ++ ExtraHeaders,
-    ResponseHeaders1 = chttpd_cors:headers(Req, ResponseHeaders),
-    {ok, MochiReq:serve_file(RelativePath, DocumentRoot, ResponseHeaders1)}.
+serve_file(Req0, RelativePath0, DocumentRoot0, ExtraHeaders) ->
```
eiri: Why is every line in this function double-spaced?

iilyak: @eiri: fixed

jaydoane:
+1

eiri:
+1

PR #151, opened 2016-03-14
Fix incorrect use of eunit fixtures

iilyak:
The tests were over-wrapped. We should return simple tests from every test case. Otherwise the tests would always pass since `eunit` wouldn't execute the test body when `?_` used in this context.

### DIFF
--- a/test/couch_db_plugin_tests.erl
+++ b/test/couch_db_plugin_tests.erl
@@ -114,77 +114,65 @@ callback_test_() ->
 
 
 validate_dbname_match() ->
-    ?_assertMatch(
-        {true, [validate_dbname, db]},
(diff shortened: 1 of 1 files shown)


eiri:
Oh, missed this one in the notifications. Test was tested, looks good : +1

On a side note - why do you prefer `?assertMatch` over `?assertEqual`, they are interchangeable here? Don't take it as a show-stopper, it's good to merge as it is, both do the job, I just wonder if I'm missing something in how `couch_db_plugin` works.

iilyak:
> why do you prefer ?assertMatch over ?assertEqual

@eiri: I used to generate random args and match that they are returned back. But then abandoned that as an overkill.

PR #153, opened 2016-03-16
Fix log_response/2

eiri:
After recent refactoring `log_response/2` can also receive an unencoded JSON as part of the error message.

### DIFF
--- a/src/couch_httpd.erl
+++ b/src/couch_httpd.erl
@@ -649,16 +649,17 @@ log_request(#httpd{mochi_req=MochiReq,peer=Peer}=Req, Code) ->
             gen_event:notify(couch_plugin, {log_request, Req, Code})
     end.
 
+log_response(Code, _) when Code < 400 ->
+    ok;
 log_response(Code, Body) ->
-    case erlang:get(dont_log_response) of
-        true ->
+    case {erlang:get(dont_log_response), Body} of
+        {true, _} ->
             ok;
-        _ when Code >= 500 ->
-            couch_log:error("httpd ~p error response:~n ~s", [Code, Body]);
-        _ when Code >= 400 ->
-            couch_log:error("httpd ~p error response:~n ~s", [Code, Body]);
+        {_, {json, JsonObj}} ->
+            ErrorMsg = couch_util:json_encode(JsonObj),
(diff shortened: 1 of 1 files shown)


iilyak:
@eiri: Thank you for noticing the problem. Wouldn't it be better to call `couch_util:json_encode(JsonObj)` after we check `dont_log_response`?

eiri:
@iilyak this sounds like micro-optimization, tbh, and leads to not the prettiest code, but if you insist -  sure, here we go.

kxepal:
+1

iilyak:
@eiri: I would disagree that it is a micro optimization. We would be doing `json -> binary` conversion for every request (which involves `{json, Obj}`) twice if it is coming from clustered interface.

src/couch_httpd.erl
```diff
@@ -649,16 +649,17 @@ log_request(#httpd{mochi_req=MochiReq,peer=Peer}=Req, Code) ->
...
-    case erlang:get(dont_log_response) of
-        true ->
+    case {erlang:get(dont_log_response), Body} of
+        {true, _} ->
             ok;
-        _ when Code >= 500 ->
-            couch_log:error("httpd ~p error response:~n ~s", [Code, Body]);
-        _ when Code >= 400 ->
-            couch_log:error("httpd ~p error response:~n ~s", [Code, Body]);
+        {_, {json, JsonObj}} ->
```
iilyak: It looks like you dropped  `400 <= Code >= 500` condition which would mean we would be logging more than before.

iilyak: Never mind.

iilyak: Never mind

iilyak:
+1 after cleaning up the history

eiri:
@iilyak I once had a production system that was doing jiffy encoding of ~1Mb messages at rate about 80 msg/s without breaking a sweet, so I have my doubts about optimizing without measuring. On the other hand if anything couch core code taught me is to treasure functions, that I can grasp propose of just glancing over, instead of stopping and reading attentively.

Anyway, everyone's entitled to their own opinions, so that's a moot point. The PR squashed.

PR #154, opened 2016-03-17
Use couch_tests applications for couch_db_plugin_tests

iilyak:
This fixes the problem with chttpd_plugin_tests. Previously following would fail:
```
make eunit apps=couch,chttpd tests=callback_test,error_info_test
```
In order to test this PR you would need to add apache/couchdb-erlang-tests#1 as a dependency in `rebar.config.script`. You might need to add the app into `rel/reltool.config` as well.


iilyak:
@kxepal: This one is the same change as in https://github.com/apache/couchdb-chttpd/pull/108 (which you already review).

kxepal:
Right. +1 here as well.

PR #156, opened 2016-03-24
Use {minor_version, 1} on term_to_binary call for MD5 of _rev id

MikeFair:
This uses the 64-bit IEEE format for floats, making the binary representation consistent between OTP versions before and after 17.0.
This also makes it easier for third parties to replicate the md5 portion of the revision id calculation.


MikeFair:
Triggering a rebuild of Travis-CI

rnewson:
First line is too long (76 chars max pls) and has a typo. Please also file a JIRA ticket and refer to it in the comment (COUCHDB-XXXX as last line).

PR #157, opened 2016-03-30
Add capability for compaction on local update

sagelywizard:
Prior to this commit, there was no functionality for a custom compactor
to know if the local database had been updated. This commit adds a
local_updated event, which will inform custom compactors when the local
tree has been updated.

### DIFF
--- a/src/couch_db_updater.erl
+++ b/src/couch_db_updater.erl
@@ -294,6 +294,10 @@ handle_info({update_docs, Client, GroupedDocs, NonRepDocs, MergeConflicts,
             couch_event:notify(Db2#db.name, updated);
         true -> ok
         end,
+        if NonRepDocs2 /= [] ->
+            couch_event:notify(Db2#db.name, local_updated);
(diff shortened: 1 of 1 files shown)


src/couch_db_updater.erl
```diff
@@ -294,6 +294,10 @@ handle_info({update_docs, Client, GroupedDocs, NonRepDocs, MergeConflicts,
             couch_event:notify(Db2#db.name, updated);
         true -> ok
         end,
+        case NonRepDocs2 of
```
rnewson: can we write this in the style of the previous notification? It's slightly shorter but, more importantly, it's consistent with the surrounding coding style.

if NonRepDocs2 /= [] ->
  couch_event:notify(...
true -> ok
end,

rnewson: and it makes the condition we care about obvious

rnewson:
+1 to the idea, I commented on the implementation style I'd prefer.

PR #158, opened 2016-03-30
Fix normalize_dbname to work with slashes

iilyak:
We shouldn't rely on dbname being the last part of path. Since database
name could include slashes.
normalize_dbname as the name suggests should accept either binary or
list and return only binary.

### DIFF
--- a/src/couch_db.erl
+++ b/src/couch_db.erl
@@ -1513,10 +1513,8 @@ select_gt(V1, _V2) -> V1.
 select_lt(V1, V2) when V1 > V2 -> V2;
 select_lt(V1, _V2) -> V1.
 
-normalize_dbname(<<"shards/", _/binary>> = Path) ->
-    lists:last(binary:split(mem3:dbname(Path), <<"/">>, [global]));
 normalize_dbname(DbName) ->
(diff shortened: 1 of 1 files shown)


eiri:
lgtm

rnewson:
hrm, is that right? is normalize_dbname/1 really exactly what mem3:dbname/1 does? (cc @davisp)

davisp:
This looks odd to me. So I have this right, an example:

Path1 == <<"shards/00000000-ffffffff/username/dbname.234234234">>
mem3:dbname(Path1) == <<"username/dbname">>
couch_db:normalize_dbname(Path1) == <<"dbname">>

We used the original normalize_dbname (though never by that name) to check if a db was a _replicator or _users db. The missing detail/assumption here was that we never cared if someone named a database like <<"foo/_replicator">> in so much as it'd just get picked up as a _replicator database. This patch and related changes appear to be wanting to avoid that when we've never cared before.

However I'd go about this fairly differently. normalize_dbname used to be this:

lists:last(binary:split(mem3:dbname(Path), <<"/">>, [global])).

I'd just remove the global option so we only split once.

PR #160, opened 2016-04-08
Change normalize_dbname semantic add dbname_suffix

iilyak:
normalize_dbname used to return last componenent of the shard path. But
it was doing so only if argument was passed as a binary. We change the
semantic of normalize_dbname so it does following:

  1. expects either type binary or list
  2. expect any kind of database name
     - plain database name
     - plain database name with '.couch' extension
     - shard name with suffix (creation time)
     - shard name with suffix (creation time) and .couch extension
  3. always returns a binary containing all components of a database name
  4. removes information about shard
  5. removes shard's suffix

For cases we need last component of the path (old behaviour) we
intoroduce dbname_suffix/1

COUCHDB-2983

### DIFF
--- a/src/couch_db.erl
+++ b/src/couch_db.erl
@@ -35,6 +35,7 @@
 -export([monitored_by/1]).
 -export([normalize_dbname/1]).
 -export([validate_dbname/1]).
+-export([dbname_suffix/1]).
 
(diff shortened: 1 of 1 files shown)


davisp:
+1

kxepal:
+1

PR #161, opened 2016-04-08
Redo `enable_database_recovery` behaviour

eiri:
All the functions related to file deletion moved into `couch_file` module.

Config option `rename_on_delete` split into two options:

1. Option  `enable_database_recovery`. When set, on user initiated deletion of database renames a database file and the database's views directories in-place with timestamp and "deleted" tag-word suffix, so the database can be recovered if necessary.
2. Option `delete_after_rename`. When set, keeps all the deleted compaction files in ".delete" directory as renamed in random uuids files until the database restart. This allows to postpone deletion of the  compaction files and provides with option of "last minute safe" in case database got corrupted after compaction.

### DIFF
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -219,12 +219,14 @@ close(Fd) ->
 
 
 delete(RootDir, Filepath) ->
-    delete(RootDir, Filepath, true).
+    delete(RootDir, Filepath, []).
 
-delete(RootDir, FullFilePath, Async) ->
+delete(RootDir, FullFilePath, Options) ->
     EnableRecovery = config:get_boolean("couchdb",
         "enable_database_recovery", false),
-    case EnableRecovery of
+    Async = not lists:member(sync, Options),
+    Context = couch_util:get_value(context, Options, compaction),
+    case Context =:= delete andalso EnableRecovery of
         true ->
             rename_file(FullFilePath);
         false ->
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -246,7 +246,11 @@ delete_file(RootDir, Filepath, Async) ->
 
 rename_file(Original) ->
     DeletedFileName = deleted_filename(Original),
-    file:rename(Original, DeletedFileName).
+    Now = calendar:local_time(),
+    case file:rename(Original, DeletedFileName) of
+        ok -> file:change_time(DeletedFileName, Now);
+        Else -> Else
+    end.
 
 deleted_filename(Original) ->
     {{Y, Mon, D}, {H, Min, S}} = calendar:universal_time(),
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -264,6 +264,18 @@ deleted_filename(Original) ->
     filename:rootname(Original) ++ Suffix.
 
 nuke_dir(RootDelDir, Dir) ->
+    EnableRecovery = config:get_boolean("couchdb",
+        "enable_database_recovery", false),
+    case EnableRecovery of
+        true ->
+            rename_file(Dir);
+        false ->
+            delete_dir(RootDelDir, Dir)
+    end.
+
+delete_dir(RootDelDir, Dir) ->
+    DeleteAfterRename = config:get_boolean("couchdb",
+        "delete_after_rename", true),
     FoldFun = fun(File) ->
         Path = Dir ++ "/" ++ File,
         case filelib:is_dir(Path) of
@@ -271,7 +283,7 @@ nuke_dir(RootDelDir, Dir) ->
                 ok = nuke_dir(RootDelDir, Path),
                 file:del_dir(Path);
             false ->
-                delete(RootDelDir, Path, false)
+                delete_file(RootDelDir, Path, false, DeleteAfterRename)
         end
     end,
     case file:list_dir(Dir) of
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -221,8 +221,16 @@ close(Fd) ->
 delete(RootDir, Filepath) ->
     delete(RootDir, Filepath, true).
 
+delete(RootDir, FullFilePath, Async) ->
+    RenameOnDelete = config:get_boolean("couchdb", "rename_on_delete", false),
+    case RenameOnDelete of
+        true ->
+            rename_file(FullFilePath);
+        false ->
+            delete_file(RootDir, FullFilePath, Async)
+    end.
 
-delete(RootDir, Filepath, Async) ->
+delete_file(RootDir, Filepath, Async) ->
     DelFile = filename:join([RootDir,".delete", ?b2l(couch_uuids:random())]),
     case file:rename(Filepath, DelFile) of
     ok ->
@@ -236,6 +244,17 @@ delete(RootDir, Filepath, Async) ->
         Error
     end.
 
+rename_file(Original) ->
+    DeletedFileName = deleted_filename(Original),
+    file:rename(Original, DeletedFileName).
+
+deleted_filename(Original) ->
+    {{Y, Mon, D}, {H, Min, S}} = calendar:universal_time(),
+    Suffix = lists:flatten(
+        io_lib:format(".~w~2.10.0B~2.10.0B."
+            ++ "~2.10.0B~2.10.0B~2.10.0B.deleted"
+            ++ filename:extension(Original), [Y, Mon, D, H, Min, S])),
+    filename:rootname(Original) ++ Suffix.
 
 nuke_dir(RootDelDir, Dir) ->
     FoldFun = fun(File) ->
@@ -610,3 +629,43 @@ process_info(Pid) ->
         {couch_file_fd, {Fd, InitialName}} ->
             {Fd, InitialName}
     end.
+
+
+-ifdef(TEST).
+-include_lib("couch/include/couch_eunit.hrl").
+
+deleted_filename_test_() ->
+    DbNames = ["dbname", "db.name", "user/dbname"],
+    Fixtures = make_filename_fixtures(DbNames),
+    lists:map(fun(Fixture) ->
+        should_create_proper_deleted_filename(Fixture)
+    end, Fixtures).
+
+should_create_proper_deleted_filename(Before) ->
+    {Before,
+    ?_test(begin
+        BeforeExtension = filename:extension(Before),
+        BeforeBasename = filename:basename(Before, BeforeExtension),
+        Re = "^" ++ BeforeBasename ++ "\.[0-9]{8}\.[0-9]{6}\.deleted\..*$",
+        After = deleted_filename(Before),
+        ?assertEqual(match,
+            re:run(filename:basename(After), Re, [{capture, none}])),
+        ?assertEqual(BeforeExtension, filename:extension(After))
+    end)}.
+
+make_filename_fixtures(DbNames) ->
+    Formats = [
+        "~s.couch",
+        ".~s_design/mrview/3133e28517e89a3e11435dd5ac4ad85a.view",
+        "shards/00000000-1fffffff/~s.1458336317.couch",
+        ".shards/00000000-1fffffff/~s.1458336317_design",
+        ".shards/00000000-1fffffff/~s.1458336317_design"
+            "/mrview/3133e28517e89a3e11435dd5ac4ad85a.view"
+    ],
+    lists:flatmap(fun(DbName) ->
+        lists:map(fun(Format) ->
+            filename:join("/srv/data", io_lib:format(Format, [DbName]))
+        end, Formats)
+    end, DbNames).
+
+-endif.
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -228,21 +228,23 @@ delete(RootDir, FullFilePath, Async) ->
         true ->
             rename_file(FullFilePath);
         false ->
-            delete_file(RootDir, FullFilePath, Async)
+            DeleteAfterRename = config:get_boolean("couchdb",
+                "delete_after_rename", true),
+            delete_file(RootDir, FullFilePath, Async, DeleteAfterRename)
     end.
 
-delete_file(RootDir, Filepath, Async) ->
+delete_file(RootDir, Filepath, Async, DeleteAfterRename) ->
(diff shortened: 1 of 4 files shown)


src/couch_file.erl
```diff
@@ -121,7 +121,7 @@ append_term_md5(Fd, Term, Options) ->
 
 append_binary(Fd, Bin) ->
     ioq:call(Fd, {append_bin, assemble_file_chunk(Bin)}, erlang:get(io_priority)).
-
```
rnewson: whitespace only change

src/couch_file.erl
```diff
@@ -223,19 +223,34 @@ delete(RootDir, Filepath) ->
...
 delete(RootDir, Filepath, Async) ->
-    DelFile = filename:join([RootDir,".delete", ?b2l(couch_uuids:random())]),
+    DelFile = deleted_filename(RootDir, Filepath),
+    RenameOnDelete = config:get_boolean("couchdb", "rename_on_delete", false),
     case file:rename(Filepath, DelFile) of
-    ok ->
-        if (Async) ->
+        ok when RenameOnDelete ->
+            Now = calendar:local_time(),
+            file:change_time(DelFile, Now);
```
rnewson: this belongs in a different commit given the description of this commit.

src/couch_file.erl
```diff
@@ -223,19 +223,34 @@ delete(RootDir, Filepath) ->
...
+            file:delete(DelFile);
+        Error ->
+            Error
     end.
 
+deleted_filename(RootDir, Filepath) ->
+    RootName = filename:rootname(Filepath),
+    DelFileRoot = couch_util:url_encode(RootName -- RootDir),
+    Suffix = deleted_filename_suffix(),
+    Extention = filename:extension(Filepath),
```
rnewson: `Extension`

src/couch_file.erl
```diff
@@ -253,14 +253,23 @@ deleted_filename_suffix() ->
...
             true ->
-                ok = nuke_dir(RootDelDir, Path),
+                ok = delete_dir(RootDelDir, Path),
                 file:del_dir(Path);
             false ->
-                delete(RootDelDir, Path, false)
+                UUID = couch_uuids:random(),
+                DelFile = filename:join([RootDelDir, ".delete", UUID]),
+                file:rename(Path, DelFile),
+                file:delete(DelFile)
```
rnewson: so it .. renames _and_ then immediately deletes? why?

src/couch_file.erl
```diff
@@ -271,7 +280,6 @@ nuke_dir(RootDelDir, Dir) ->
             ok
     end.
 
-
```
rnewson: ws only change

src/couch_server.erl
```diff
@@ -555,7 +555,12 @@ delete_file(RootDir, FullFilePath, Options) ->
 rename_on_delete(Original) ->
     DeletedFileName = deleted_filename(Original),
     case file:rename(Original, DeletedFileName) of
-        ok -> {ok, {renamed, DeletedFileName}};
+        ok ->
```
rnewson: parts of this change seem to be in the first commit (a file:change_time call is in there)

src/couch_file.erl
```diff
@@ -223,29 +223,53 @@ delete(RootDir, Filepath) ->
...
+
+deleted_filename_suffix() ->
+    {{Y, Mon, D}, {H, Min, S}} = calendar:universal_time(),
+    io_lib:format("~w~2.10.0B~2.10.0B"
+        ".~2.10.0B~2.10.0B~2.10.0B.~6s",
+        [Y, Mon, D, H, Min, S, couch_uuids:random()]).
 
 nuke_dir(RootDelDir, Dir) ->
+    case config:get_boolean("couchdb", "rename_on_delete", false) of
+        true -> couch_server:delete_file(RootDelDir, Dir, []);
```
iilyak: I don't like the fact that we introduce cyclic dependency between couch_file and couch_server here. But I wouldn't insist on changing it. Feel free to ignore.

src/couch_file.erl
eiri: I believe this is a performance thing related to Windows fs, that why `couch_file:delete/3` does the same thing - renames and then deletes. Since it was called here previously with `sync`, I'm mimic it, but with simplified renaming approach to avoid taxing calls for `config`, `calendar` and `url_encode`.

iilyak: Should we `s/DelFile/TmpFile/` to avoid confusion?

eiri: Here is JIRA issue that gives more details on rename-then-delete approach: https://issues.apache.org/jira/browse/COUCHDB-1003

eiri:
@rnewson @iilyak - the comments addressed.

Regarding rename-then-delete approach, just to move the comment here out of the outdated diff:

As described here https://issues.apache.org/jira/browse/COUCHDB-1003 , this is done because deletion of large files on block-oriented filesystems can take time, so the file first moved into `.deleted` directory that got cleaned up on couch start.

@rnewson if you think this is not applicable here, I'll remove the rename part, I was just reproducing behaviour of `nuke_dir/2` as it was before this PR.

rnewson:
@eiri I'm still confused. The way Cloudant addresses the problem of deletes taking a long time is that we do _not_ call file:delete on it. Instead, we truncate it a few meg at a time until it's 0 length then delete it. Your code does a rename then a full delete, it doesn't achieve the goal.

kxepal:
@rnewson Just curious: what the idea stands behind "truncate by few MiBs toward 0 and then delete" and why it's better than "rename and full delete"? The latter sounds to be use less disk I/O operations and be better.

eiri:
@rnewson My goal was not change the delete process, but to preserve existing behaviour and the code does _exactly_ what've been done here previously. My understanding from your comment in linked jira is that it was done this way, so if Couch goes down during a long delete the file still be purged on a restart.

iilyak:
I'm +1 with the change. However I want a +1 from @rnewson on this one before we merge.

rnewson:
@kxepal: because the kernel/filesystem will devote itself to freeing the blocks, it causes all other reads/writes to suffer hugely. It _is_ more expensive to repeatedly truncate vs simple delete, but it's spaced out over time so affects other operations considerably less.

rnewson:
@eiri the first PR comment says 'This change makes compaction files to stay in .deleted directory in case config parameter rename_on_delete is set and renames them in fashion, that makes it easier to restore them if necessary. ' but what you've done is rename-then-delete which does not achieve the goal.

I think rename_on_delete might be a misleading name. rename_instead_of_delete is the meaning. Wherever we delete a file, whether that's a database or view compaction that completes or a user deletes a database, we want one of two behaviours: 1) we delete the file immediately (yes, we'll mv it to the .delete directory first so that we can clean up easily in the event of a crash, but that's normal couchdb behaviour) or 2) we rename the file (appending a unique timestamp or similar) and do not delete it, so that it can be recovered. Anyone that sets this config flag is then required to find and delete these files later if they don't intend to restore them (so a useful pattern like *.deleted.couch / *.deleted.view is warranted).

Have I misunderstood the scope of this pull request?

kxepal:
@rnewson Interesting. What would you recommend to read about? (sorry for offtopic)

rnewson:
@kxepal it was an observation from production but I'm sure you can trivially reproduce it. make a 200gb file on ext4, delete it while running any kind of benchmark.

davisp:
@kxepal We actually stole the truncate-to-zero approach from a talk at one of the devops conferences some years ago. Its purely based on the observation that @rnewson made, when you delete a file, the kernel will consume a large amount of disk resources trying to remove that file. Seeing it in the talk made us realize why we ended up having such a performance impact when a compaction finished. Here's a couple quick links that describe similar solutions:

http://www.awasu.com/weblog/deleting-large-files-from-an-ext3-file-system/
http://serverfault.com/questions/480526/deleting-very-large-file-without-webserver-freezing

The ionice suggestion in the second link I'm pretty sure depends on having a kernel io scheduler configured and we've seen that have a performance impact with anything other than the noop scheduler.

davisp:
I should've made the point that ionice doesn't work if you're using noop io scheduling.

kxepal:
@rnewson @davisp Thanks a lot!

kxepal:
Hm...reading the links, references and quick googling shows that this problem is extX specific and other FS may handle delete right without IOpain. Feels like this algorithm better to make configurable and notify admins about such problem and the option, but I would like to make some benchmarks for FS I uses for CouchDB: XFS and ZFS. Though for XFS this [shouldn't be a problem](http://www.bullopensource.org/ext4/20070404/FileDeletion.html) already.

eiri:
@rnewson View compaction files handled in `couch_file:deleted/3` function which does what the first PR comment describes and you are commenting on `couch_file:delete_dir/2` which is called on database's views directories when database deleted while parameter `rename_on_delete` set to false [[ref]](https://github.com/cloudant/couchdb-couch/blob/61028-fix-rename-on-delete/src/couch_file.erl#L258)

The scope of PR is to extend exiting `rename_on_delete` behaviour on compaction and view files. I'm doing this by making views directories to be renamed on deletion of a database in the same manner the database file renamed right now. For compaction files I'm making them not to be deleted in `.deleted` directory after renaming, the same way as Cloudant does. I'm renaming them differently, though, so they'd be more easily identified if necessary.

The renaming configuration parameters or radically diverging from how file preservation works right now is not in scope of this PR.

rnewson:
(noting that ionice doesn't apply to delete operations anyway)

src/couch_file.erl
```diff
@@ -223,19 +223,33 @@ delete(RootDir, Filepath) ->
...
+            ok;
+        ok when Async ->
             spawn(file, delete, [DelFile]),
             ok;
-        true ->
-            file:delete(DelFile)
-        end;
-    Error ->
-        Error
+        ok ->
```
rnewson: diff would have been clearer if you hadn't mixed whitespace changes with functional changes.

src/couch_file.erl
```diff
@@ -252,14 +252,23 @@ deleted_filename_suffix() ->
...
             true ->
-                ok = nuke_dir(RootDelDir, Path),
+                ok = delete_dir(RootDelDir, Path),
                 file:del_dir(Path);
             false ->
-                delete(RootDelDir, Path, false)
+                UUID = couch_uuids:random(),
+                TmpFile = filename:join([RootDelDir, ".delete", UUID]),
+                file:rename(Path, TmpFile),
+                file:delete(TmpFile)
```
rnewson: this still does seem a bit pointless in this context. The original context was deleting a database which, at the time, was a single file. When we're deleting all views, renaming each filed them immediately deleting it doesn't seem to do much beyond complicate things.

@davisp thoughts here pls?

I suggest that nuke_dir renames the top level Dir to .delete first and then recursively deletes it. That gives us the behaviour we want (making it easier to clean if we crash and restart during the time-consuming file:delete calls)

davisp: I agree. The rename and delete isn't very useful here. I'd do what @rnewson suggests and just rename in nuke_dir, and then do a similar test as above in couch_file:delete that if rename_on_delete is true then we just don't do the recursion similar to what exists currently.

davisp:
So thinking through this I think we may be conflating two different features that while similar are serving different purposes.

Specifically, there's one feature where we rename things into the .delete directory so that we can get rid of them quickly as well as manage cleanup when the server crashes during deletions. This feature has been in CouchDB for a long time dating back years.

The second feature was a small thing that Cloudant added to help manage recovering from accidental deletions. This feature was implemented by just changing the delete to rename a .couch file to a version that includes the deletion timestamp. This allowed operators to recover a "deleted" database for clients.

This PR appears to attempt to flatten both of these features by renaming files into the .delete directory with useful names and makes the actual deletion a bit subtler in that the various renames are munged and so its not 100% obvious which feature we're handling in the various functions named delete/nuke_dir/delete_file/etc.

As a follow on, the renaming here also tries to move away from using a UUID in the .delete directory and instead uses a URL encoded path of the deleted file (relative to the root database directory). On first glance this seemed like a good idea, but researching filename length limits it'd be very easy to break this as a user only needs to create a database with > 255 characters on extX filesystems (and shorter on others).

What I think we should consider is that all of our deletions move the file to the .delete directory mirroring the original filesystem hierarchy. Then our existing "rename_on_delete" feature is relabeled as "delete_after_rename" with a default of true. This allows us to enable a sysadmin approach to recovering deleted databases as well as enables sysadmins to institute their own policies for actual deletion.

As a last note, I'd also like to see all of the file system/deletion logic moved to couch_file. Its quite awkward to have it split between couch_server and couch_file. I'd work on structuring the commits to have one that moves the existing logic to couch_file and then a subsequent commits to do the filesystem hierarchy mirroring.

eiri:
@davisp I thought there are performance penalty on deleting a lot of files in deep filesystem hierarchy, I remember @rnewson mentioned this while we've been discussing how `rename_on_delete` should be done before I started this implementation. That's why I did urlencode trick to keep the file structure flat.

Do I understand correctly that you are proposing instead: 1) to change `rename_on_delete` config parameter for `delete_after_rename` 2) keep doing what we are doing with renaming files in uuids, moving them in flat fashion in `.delete` and removing them if `delete_after_rename` set to `true`. 3) mirroring data directory hierarchy in `.delete` directory, moving there after-compaction, deleted database and view files and disabling `init_delete_dir/1` if `delete_after_rename` set to `false`

If this is the case, do you think we should add random suffix to moved after-compaction files to keep them distinguished? On one hand this could lead to accumulation of a lot of files in `.delete`, on a flip hand rapid compaction without suffixing would override previous compactions files that might have some value.
Should we change config parameter to `rename_instead_of_delete` as @rnewson proposed, since there are two different ways to rename files and strictly speaking Couch admin don't necessary need to know how we are deleting files when we are deleting them for real.
Switching `rename_instead_of_delete` (or `delete_after_rename`) during Couch life could lead to an interesting state in `.delete` directory - do you think this is not a concern?

eiri:
Companion PR apache/couchdb-couch-mrview#44

iilyak:
Great work @eiri. All tests pass locally.

eiri:
@davisp I've redone this "database recovery" thing in a fashion we discussed, can you please take a look when you'll have some time? thank you.

davisp:
Awesome work. +1

PR #162, opened 2016-04-15
Implement Mango selectors for change feeds

nickva:
API is modeled after _doc_ids filter for change feeds.

User POSTs to {db}/_changes with `filter=_selector`.
Document body should have a "selector" field, with
a Mango selector object as value.

For example:
```
http://.../d1/_changes?filter=_selector
{
  "selector": {"z" : {"$gte" : 1} }
}
```
In case "selector" is missing or its value is not a
JSON object, a 400 error is returned with a
corresponding error message. For example:
```
{
    "error": "bad_request",
    "reason": "Selector error: 42, expected an object"
}
```

In case "selector" object cannot be validated as a
Mango selector, a 400 error is returned with a validation
error as "reason".

Jira: COUCHDB-2988

### DIFF
--- a/src/couch_changes.erl
+++ b/src/couch_changes.erl
@@ -196,6 +196,8 @@ get_callback_acc(Callback) when is_function(Callback, 2) ->
 
 configure_filter("_doc_ids", Style, Req, _Db) ->
     {doc_ids, Style, get_doc_ids(Req)};
+configure_filter("_selector", Style, Req, _Db) ->
+    {selector, Style,  get_selector(Req)};
 configure_filter("_design", Style, _Req, _Db) ->
     {design_docs, Style};
 configure_filter("_view", Style, Req, Db) ->
@@ -267,6 +269,11 @@ filter(_Db, DocInfo, {doc_ids, Style, DocIds}) ->
         false ->
             []
     end;
+filter(Db, DocInfo, {selector, Style, Selector}) ->
+    Docs = open_revs(Db, DocInfo, Style),
+    Passes = [mango_selector:match(Selector, couch_doc:to_json_obj(Doc, []))
+        || Doc <- Docs],
+    filter_revs(Passes, Docs);
 filter(_Db, DocInfo, {design_docs, Style}) ->
     case DocInfo#doc_info.id of
         <<"_design", _/binary>> ->
@@ -336,6 +343,15 @@ get_doc_ids(_) ->
     throw({bad_request, no_doc_ids_provided}).
 
 
+get_selector({json_req, {Props}}) ->
+    check_selector(couch_util:get_value(<<"selector">>, Props));
+get_selector(#httpd{method='POST'}=Req) ->
+    couch_httpd:validate_ctype(Req, "application/json"),
+    get_selector({json_req,  couch_httpd:json_body_obj(Req)});
+get_selector(_) ->
+    throw({bad_request, "Selector must be specified in POST payload"}).
+
+
 check_docids(DocIds) when is_list(DocIds) ->
     lists:foreach(fun
         (DocId) when not is_binary(DocId) ->
@@ -347,6 +363,18 @@ check_docids(DocIds) when is_list(DocIds) ->
 check_docids(_) ->
     Msg = "`doc_ids` filter parameter is not a list of doc ids.",
     throw({bad_request, Msg}).
+
+
+check_selector(Selector={_}) ->
+    try
+        mango_selector:normalize(Selector)
+    catch
+        {mango_error, Mod, Reason0} ->
+            {_StatusCode, _Error, Reason} = mango_error:info(Mod, Reason0),
+            throw({bad_request, Reason})
+    end;
+check_selector(_Selector) ->
+    throw({bad_request, "Selector error: expected a JSON object"}).
 
 
 open_ddoc(#db{name=DbName, id_tree=undefined}, DDocId) ->
--- a/test/couch_changes_tests.erl
+++ b/test/couch_changes_tests.erl
@@ -60,6 +60,7 @@ changes_test_() ->
             setup,
             fun test_util:start_couch/0, fun test_util:stop_couch/1,
             [
+                filter_by_selector(),
                 filter_by_doc_id(),
                 filter_by_design(),
                 continuous_feed()
@@ -83,6 +84,24 @@ filter_by_doc_id() ->
             ]
         }
     }.
+
+filter_by_selector() ->
+    {
+        "Filter _selector",
+        {
+            foreach,
+            fun setup/0, fun teardown/1,
+            [
+                fun should_select_basic/1,
+                fun should_select_with_since/1,
+                fun should_select_when_no_result/1,
+                fun should_select_with_deleted_docs/1,
+                fun should_select_with_continuous/1,
+                fun should_stop_selector_when_db_deleted/1
+            ]
+        }
+    }.
+
 
 filter_by_design() ->
     {
@@ -317,7 +336,7 @@ should_filter_continuous_feed_by_specific_doc_ids({DbName, Revs}) ->
 
 should_end_changes_when_db_deleted({DbName, _Revs}) ->
     ?_test(begin
-        {ok, Db} = couch_db:open_int(DbName, []),
+        {ok, _Db} = couch_db:open_int(DbName, []),
         ChangesArgs = #changes_args{
             filter = "_doc_ids",
             feed = "continuous"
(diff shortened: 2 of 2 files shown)


src/couch_changes.erl
```diff
@@ -336,6 +343,16 @@ get_doc_ids(_) ->
     throw({bad_request, no_doc_ids_provided}).
 
 
+get_selector({json_req, {Props}}) ->
+    check_selector(couch_util:get_value(<<"selector">>, Props));
+get_selector(#httpd{method='POST'}=Req) ->
+    couch_httpd:validate_ctype(Req, "application/json"),
+    {Props} = couch_httpd:json_body_obj(Req),
+    check_selector(couch_util:get_value(<<"selector">>, Props));
```
kxepal: Seems like you repeat yourself here. How about to change that branch into:
```
get_selector(#httpd{method='POST'}=Req) ->
   couch_httpd:validate_ctype(Req, "application/json"),
   get_selector({json_req, couch_httpd:json_body_obj(Req)});
```

You'll also produce bad request instead of server internal error with function clause in case of invalid request json data, like array.

src/couch_changes.erl
```diff
@@ -336,6 +343,16 @@ get_doc_ids(_) ->
...
+get_selector(_) ->
+    throw({bad_request, invalid_selector}).
```
kxepal: Probably, `invalid_selection` is misleading as it actually about invalid request data. All about invalid selector should throw check_selector fun.

src/couch_changes.erl
```diff
@@ -349,6 +366,19 @@ check_docids(_) ->
...
+check_selector(Selector={_}) ->
+    try
+        mango_selector:normalize(Selector)
+    catch
+        {mango_error, _Mod, Reason} -> % see MANGO_ERROR(R) in mango.hrl
+            Msg = io_lib:format("Selector error: ~p", [Reason]),
+            throw({bad_request, Msg})
+    end;
+check_selector(Selector) ->
+    Msg = io_lib:format("Selector error: ~p, expected an object", [Selector]),
```
kxepal: May be not send back the Selector data? The case: send ~1GiB of data like `{"selector": [....]}` and let server try to serialize that 1GiB of data and send it back to you.

kxepal:
Nice feature, but I feel quite strange about creating dependency from mango for couch app. Since mango is a cluster thing, shouldn't it be on chttpd side?

src/couch_changes.erl
nickva: Tried that change but now a simple selector: {} which worked before fails with:
```
-    {Props} = couch_httpd:json_body_obj(Req),
-    check_selector(couch_util:get_value(<<"selector">>, Props));
+    check_selector({json_req, couch_httpd:json_body_obj(Req)});
```
{ ... "reason": "Selector error: {json_req,{[{<<\"selector\">>,{[]}}]}}, expected an object"...}

kxepal: Last line: `get_selector`, not `check_selector`. Also, point about function_clause is false: I miss the moment that json_body_obj already handles all the routines to ensure that we got JSON object, nothing else.

src/couch_changes.erl
nickva: It is an invalid_selector because this check happens after noticing there is a _selector as filter. Which prompted the the call to `configure_filter("_selector",...`. At this point if it is not  GET request, or object not is an array it is an "invalid selector".

src/couch_changes.erl
nickva: Agreed. Good point!

src/couch_changes.erl
kxepal: In case of bad selector type - yes, but in case of GET request it's just missed. May be add one more clause about:
```
get_selector({json_req, _}) -> throw({bad_request, invalid_selector});
```

And the this clause turns into something more friendly:
```
get_selector(_) -> throw({bad_request, "Selector must be specified in POST payload"}).
```

What do you think?

src/couch_changes.erl
```diff
@@ -349,6 +366,18 @@ check_docids(_) ->
     throw({bad_request, Msg}).
 
 
+check_selector(Selector={_}) ->
+    try
+        mango_selector:normalize(Selector)
+    catch
+        {mango_error, _Mod, Reason} -> % see MANGO_ERROR(R) in mango.hrl
+            Msg = io_lib:format("Selector error: ~p", [Reason]),
```
davisp: You want `mango_error:info(Mod, Reason) which gives you back a {StatusCode, Error, Reason} triple where Reason is formatted correctly.

src/couch_changes.erl
davisp: The pattern that @kxepal is highlighting is to recurse to unwrap objects. So your first clause has the base case, and the second clause would take out the thing to be selected and then recurse to the same check. That way your extraction logic is only written once.

src/couch_changes.erl
davisp: I agree with @kxepal here, if a user does a GET with a filter=_selector, then it'll be super opaque as to what wen't wrong. Though rather than a 400 you might throw a whatever it is for invalid method. Though i'd want to hear from @rnewson or someone if restricting methods based on query string paramters is kosher. If not kosher, than at least the 400 with the "you must specify the selector in a POST body" message would be fine so that users have a bread crumb when they get the error.

davisp:
@kxepal for the clustered vs. not dependency, it is slightly odd. However the changes code is sadly tied into the main couch repo, there's nothing we can really do to undo that without some pretty awkward contortions just for purity's sake.

Also, in this case, the mango logic being used is purely functional, there's no cluster or anything. Its strictly just a "does this json blob match this other json blob using the mango selector DSL rules" function.

So while I do agree it would be best to have this specific change in a different app, I'd argue that it's more of that all of this particular changes code should be in a different app that understands clustered vs not rather than the hodge podge of things we have now. Until then I wouldn't worry about this too much since it will work for both clustered and single node.

davisp:
Except for the various tweaks to error messaging, +1 to the feature. Sure did come out a lot simpler than I was expecting.

src/couch_changes.erl
rnewson: The spec (obviously) is not so clear. We could send 405 Method Not Allowed but that's a resource-level thing and seems inappropriate. I'd send a 400 with a clear message if ?selector param exists, as both kxepal and davisp suggest.

src/couch_changes.erl
nickva: Ah sorry. Missed that. Will fix it.

src/couch_changes.erl
nickva: Ok, good point. I can make that change.

kxepal:
@davisp Ok, thanks. Indeed decoupling these things would be quite problematic task now without any significant profit.

rnewson:
impressive brevity for such a powerful new feature.

nickva:
@kxepal @davisp  @rnewson  Updated error message for bad methods to be explicit, fixed get_selector duplicate code, use mango_error:info to format selector error messages.

kxepal:
+1

davisp:
Good point @rnewson. I'd go with the 400 approach then.

tonysun83:
simpler than I thought, +1

colinskow:
That's a great feature! In order to maintain a live list of docs that match a query, I will need changes from docs that used to match the query but no longer do so I can remove them from the list. Is that possible with this feature?

nickva:
@colinskow there is no easy way to add that I can think of. The query is in the context of the changes request. As long as there are results filtering back from that query, they'll contain documents which match the selector. There is nothing returning from the query which knows which documents matched before (using a previous changes request).

colinskow:
@nickva thank you very much for your response. A "live query" would be a very useful feature for front-end app developers.

I have [successfully implemented this](https://github.com/colinskow/pouchdb-live-find), but it requires listening to the entire changes feed. So my solution would not be entirely practical with a noisy database.

Would it be practical to maintain state of the query in a `_design`, `_replication`, or `_live_query` doc so the front end can maintain a live list of docs that match that query without having to listen to other noise?

nickva:
@colinskow that implies some kind of persistence. The selector passed in a change feed request is active only in the context of the POST request. There could be multiple such concurrent requests. After request is finished, there is no history preserved of previous request.

Or were you thinking about a long / continuous request, and while it is returning data back to the user,  a document which passed the filter and was returned before, was updated to no longer pass the selector. But there is nothing to tell the user about it. So perhaps have something in the change feed saying "this document passed the selector before but no longer does".

That is an interesting case then. If document is deleted we return it with `"_deleted":true`, is it reasonable to have something like `"_selector_failed":true`? I am not sure about this. You might want to take this up on the couchdb mailing list. This looks like an API change then and needs a better discussion.

PR #168, opened 2016-04-27
Undo our character advancement when we fixup

iilyak:
We should undo our character advancement after we finish fixup.

COUCHDB-1425


davisp:
+1

PR #169, opened 2016-05-03
Raise better exception on decompress of corrupt binary

eiri:
Right now when a binary without compression prefix passed on `couch_compress:decompress/1` (for example if a database file got corrupted), the module throws `function_clause` exception.

This patch adds tests for `couch_compress` module and makes `decompress/1` and `is_compressed/2` functions to raise `invalid_compression` error when they can't detect compression of passed binary.


iilyak:
+1

iilyak:
All tests pass locally.

janl:
“tests pas locally” are not a great reason to ignore CI builds. I want to merge another thing and now I am blocked. We should never do this.

This seemed to have happened here: https://github.com/apache/couchdb-couch/commit/78f575ec2cac823605561f1b38e0761d8170ac5f#commitcomment-17480742  so I’m wondering how you manage to get no local build fails without this (unless you hadn’t rebased against that before).

PR #173, opened 2016-05-16
Add config parameter to limit maximum pread size

eiri:
When a database file goes corrupt it is possible that pread will receive bogus length and place a large part of the file in memory before couch_file process will crash. Depending on a size of the file this could bring the whole node down.

This change prevents pread from reading beyond end of a file and adding configuration parameter `max_pread_size` to give a method to address a situation when corrupted read does not go beyond eof, but still makes starting node crash on a large read.

Two stats counters been added to report both exceptions.

### DIFF
--- a/priv/stats_descriptions.cfg
+++ b/priv/stats_descriptions.cfg
@@ -218,3 +218,11 @@
     {type, histogram},
     {desc, <<"duration of validate_doc_update function calls">>}
 ]}.
+{[pread, exceed_eof], [
+    {type, counter},
+    {desc, <<"number of the attempts to read beyond end of db file">>}
+]}.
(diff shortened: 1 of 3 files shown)


kxepal:
Nice!

iilyak:
Tests passing locally
```
> make eunit apps=couch tests=read_write_test,pread_limit_test,header_test,delete_test,nuke_dir_test,open_close_test
...
All 69 tests passed.
```

iilyak:
I'm +1 for this change. One more reviewer would be good though.

eiri:
@davisp Can you take a quick look, please?

davisp:
+1

PR #174, opened 2016-05-17
Support create_if_missing option in couch_db:open

iilyak:
Pass `create_if_missing` option to be used in cases like these:

- [fabric_rpc:get_or_create_db/2](https://github.com/apache/couchdb-fabric/blob/master/src/fabric_rpc.erl#L270)
- [mem3_rpc:get_or_create_db/2](https://github.com/apache/couchdb-mem3/pull/21/files#diff-4d22d90c9d6ea299758d2f1262352379R278)


rnewson:
+1

PR #176, opened 2016-05-25
Fix some type errors

iilyak:
Update couple of type specs.

Fix couple of real bugs in:
- `couch_event_sup`
- `couch_proc_manager`


nickva:
+1. Nice find.  But wonder what is happening with travis CI.

iilyak:
I hope this https://github.com/apache/couchdb/pull/420 would fix CI issues.

PR #177, opened 2016-06-02
Add optional `fields` to change feed selectors

nickva:
(This could be postponed until after 2.0 release. Issued to get comments and review).

When using selectors with `include_docs=true` can specify an optional fields
array in the POST request JSON body.

Each element in the array can be a json field (or even a key path
specified as field1.field2...). Resulting documents will contain only the
specified document fields.

For example:

```
http://.../d1/_changes?filter=_selector&include_docs=true

{
  "selector": {"z" : {"$gte" : 1} }, "fields": ["field1", "field2"]

}
```

Will first select only document with "z" value >= 1, then will return only field1 and field2 in documents.

`{ "field1": "field1value", "field2": "field2value"}`

(This requires a companion pr in fabric to work)

Jira: COUCHDB-2988

### DIFF
--- a/src/couch_changes.erl
+++ b/src/couch_changes.erl
@@ -197,7 +197,7 @@ get_callback_acc(Callback) when is_function(Callback, 2) ->
 configure_filter("_doc_ids", Style, Req, _Db) ->
     {doc_ids, Style, get_doc_ids(Req)};
 configure_filter("_selector", Style, Req, _Db) ->
-    {selector, Style,  get_selector(Req)};
+    {selector, Style,  get_selector_and_fields(Req)};
 configure_filter("_design", Style, _Req, _Db) ->
     {design_docs, Style};
 configure_filter("_view", Style, Req, Db) ->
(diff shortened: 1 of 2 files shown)


mayya-sharipova:
Have done some testing including on nested fields: everything works well as expected.

nickva:
Related to: https://github.com/apache/couchdb-fabric/pull/55

src/couch_changes.erl
```diff
@@ -377,6 +379,21 @@ check_selector(_Selector) ->
...
+check_fields(nil) ->
+    nil;
+check_fields(Fields) when is_list(Fields) ->
+    try
+        {ok, Fields1} = mango_fields:new(Fields),
+        Fields1
+    catch
+        {mango_error, Mod, Reason0} ->
+            {_StatusCode, _Error, Reason} = mango_error:info(Mod, Reason0),
+            throw({bad_request, Reason})
```
kxepal: Doesn't mango have own function to throw right bad request in case of something like that? I think I recall something like that there...

src/couch_changes.erl
```diff
@@ -377,6 +379,21 @@ check_selector(_Selector) ->
...
+    end;
+check_fields(_Fields) ->
+    throw({bad_request, "Selector error: fields must be JSON array"}).
```
kxepal: Doesn't mango has own validators for that kind of errors?

kxepal:
LGFM

src/couch_changes.erl
nickva: It does, it throws `{mango_error,...` but in this context other validation functions return `{bad_request,...` so made mango fields parsing do the same

tonysun83:
+1 on the source. maybe add an extra test with one valid field and one invalid field, u'll only get the valid field back.

nickva:
@tonysun83 good idea, will do

PR #181, opened 2016-06-24
Fix changed feed view filter for clustered access

eiri:
A changes feed can be filtered using view's btree instead of a database btree, a feature known as `fast_view`. This feature not implemented on a clustered interface, however view filter automatically marked as `fast_view` when ddoc has option `seq_indexed` enabled, which leads to a crash of filtered replication.

This change allows `fast_view` filter to be processed as a regular view filter when used on a clustered
interface.

This is a temporary fix for 2.0 release made in consideration of feature freeze. It'll be revisited after the release after `fast_view` will be added into fabric.

COUCHDB-3044

### DIFF
--- a/src/couch_changes.erl
+++ b/src/couch_changes.erl
@@ -281,7 +281,8 @@ filter(_Db, DocInfo, {design_docs, Style}) ->
         _ ->
             []
     end;
-filter(Db, DocInfo, {view, Style, DDoc, VName}) ->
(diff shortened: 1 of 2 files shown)


kxepal:
+1

eiri:
@davisp here is the short fix we've discussed on Monday.

iilyak:
+1

PR #182, opened 2016-07-06
Remove maybe_upgrade_password_hash

iilyak:
Starting from 411c130ea63c3423d7ea71743095dfb446680da6 we accidentially
upgrading user's password to pbkdf.
Removing upgrade code.


rnewson:
+1

PR #184, opened 2016-07-13
Skip couchdb_os_daemons_tests on Windows

wohali:
There are issues launching the .escript files automatically on Windows,
and the .sh files in the test will never run on Windows correctly. In the
interest of speeding along our release I'd like to simply skip these tests
on Windows for now. If launching a daemon completely fails we'll know
via the mrview tests.

Addresses COUCHDB-3040 but does not close it.


kxepal:
+1

PR #186, opened 2016-07-17
Replace `stale` with `update` in compression tests

banjiewen:
COUCHDB-3063


banjiewen:
See https://github.com/apache/couchdb-couch-mrview/pull/51.

PR #187, opened 2016-07-18
use couch epi decide

iilyak:
It is impossible to make the validation policy more tight from a plugin. This commit would let the plugin to do either tighten up or loosen the validation policy.  

COUCHDB-3066


kxepal:
+1

P.S. What is 70457 is?

rnewson:
+1 but needs a JIRA ticket.

rnewson:
@kxepal that's the ticket number from Cloudant's fogbugz tracker. We shouldn't have used it, sorry. We'll get better at using JIRA exclusively once 2.0 is out and we're all working mainly on couch repos.

kxepal:
@rnewson Soon...thanks! (:

PR #188, opened 2016-07-25
fix validate dbname

iilyak:
Fix the case when DbName contains path to db file
In the case when DbName contains the file name we need to remove .couch
extension in order to match ?DBNAME_REGEX.

[COUCHDB-3080](https://issues.apache.org/jira/browse/COUCHDB-3080)


davisp:
+1

PR #191, opened 2016-08-16
Simplify proc manager by moving assignment logic in one place

eiri:
This changes how proc manager handles proc assignment. Instead of doing this in three different places: `get_proc` call handler, `return_proc/2` with `maybe_assign_proc` and `flush_waiters/2`, proc manager now just places all the incoming requests in the waiting queue and then flushes it. 

As a result all the logic kept in one place which makes it more obvious that we are treating proc management as a processing of a single FIFO queue with "soft" and "hard" upper limits.

Consequently this is fixing a bug in `maybe_assign_proc` where it was possible to assign a client a process that wasn't aware of it.

COUCHDB-3095


eiri:
Reference on conversation about this change #190

eiri:
@davisp Here is the redo we talked about.

davisp:
Awesome work. +1

PR #192, opened 2016-08-19
Update handle_config_terminate API

iilyak:
Update handle_config_terminate API to match changes in https://github.com/apache/couchdb-config/pull/11

COUCHDB-3102

### DIFF
--- a/src/couch_external_server.erl
+++ b/src/couch_external_server.erl
@@ -20,6 +20,7 @@
 -include_lib("couch/include/couch_db.hrl").
 
 -define(RELISTEN_DELAY, 5000).
+-define(CONFIG_SUBSCRIPTION, [{"couchdb", "os_process_timeout"}]).
 
 % External API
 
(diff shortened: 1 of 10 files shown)


src/couch_external_server.erl
```diff
@@ -40,47 +38,52 @@ execute(Pid, JsonReq) ->
...
     couch_log:info("EXTERNAL: Starting process for: ~s", [Name]),
     couch_log:info("COMMAND: ~s", [Command]),
+    ok = config:subscribe_for_changes([{"couchdb", "os_process_timeout"}]),
     process_flag(trap_exit, true),
     Timeout = list_to_integer(config:get("couchdb", "os_process_timeout",
         "5000")),
     {ok, Pid} = couch_os_process:start_link(Command, [{timeout, Timeout}]),
-    ok = config:listen_for_changes(?MODULE, Pid),
-    {ok, {Name, Command, Pid}}.
+    {ok, {Name, Command, Pid, whereis(config_event)}}.
```
davisp: Seems like a lot of extra work just for that one 'EXIT' clause. I do see that without it we'd enter that clause for any 'EXIT' rather than throwing a function clause error. Though given that its rare I'd probably just undo the state tuple change and if we get an unknown 'EXIT' message we just erlang:error on it.

davisp:
Just had the one trivial comment. If you prefer the existing approach feel free to ignore my suggestion.

+1

src/couch_external_server.erl
iilyak: > @davisp:  and if we get an unknown 'EXIT' message we just erlang:error on it.
We need config_event Pid to be able to distinguish among EXIT messages.

Rather then updating state I also considered keep Pid in the process dictionary. Let me know if you like it better.

davisp: It occurs to me that my idea is dumb. By the time we get the EXIT message thr whereis will either fail or give a new pid. I think state is best like you have it already.

PR #193, opened 2016-09-01
Fix validation for ddoc and _local id

eiri:
Because binary match treats end of binary as `<<>>` we are validating "_design/" and "_local/" as proper ids.

COUCHDB-3114

### DIFF
--- a/src/couch_doc.erl
+++ b/src/couch_doc.erl
@@ -169,6 +169,10 @@ parse_revs(_) ->
 
 validate_docid(<<"">>) ->
     throw({illegal_docid, <<"Document id must not be empty">>});
+validate_docid(<<"_design/">>) ->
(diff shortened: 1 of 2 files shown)


kxepal:
+1

rnewson:
hm, there must be a better way. we now have two bits of special logic in the same function for the same reason.

rnewson:
try
```erlang
 <<"_design/", Rest/binary>> when byte_size(Rest) > 0 -> ok
```

That way we only say the _design and _local rule once each. anything else beginning _ will hit the general prohibition clause.

eiri:
@rnewson I did it at first, but then decided to go with less confusing error message for the users sake. If you think it's not a problem here then sure, I'll use guards.

rnewson:
hrm, fair enough. I can't see a tidier way to say it once but still give good error messages.

+1.

PR #194, opened 2016-09-02
Trigger cookie renewal on /_session

adrienverge:
It is currently not possible to programmatically renew an `AuthSession` cookie. The only way is to wait 10% of the cookie expiration time then make any request to CouchDB, so it sends a new cookie.

Programmatically getting cookies can be needed, for example, when passwords are managed by authentication server. E.g., a third-party auth server gets a cookie, forwards it to a user, and the user needs the CouchDB server to install a new cookie in his/her browser with a `Set-Cookie` header. (Browsers don't allow JS to install cookies, only a `Set-Cookie` header from the server can set the cookie.)

This patch brings the following change: an already-authenticated request (either POST or GET) on `/_session` will trigger the creation of a new cookie, whatever we're inside the 10% timeout window or not.  This way, applications can programmatically get a cookie without having to wait 60 seconds.

This closes: COUCHDB-3125

### DIFF
--- a/src/couch_httpd_auth.erl
+++ b/src/couch_httpd_auth.erl
@@ -223,13 +223,15 @@ cookie_authentication_handler(#httpd{mochi_req=MochiReq}=Req, AuthModule) ->
                     TimeStamp when CurrentTime < TimeStamp + Timeout ->
                         case couch_passwords:verify(ExpectedHash, Hash) of
                             true ->
-                                TimeLeft = TimeStamp + Timeout - CurrentTime,
                                 couch_log:debug("Successful cookie auth as: ~p",
                                                 [User]),
+                                TimeLeft = TimeStamp + Timeout - CurrentTime,
+                                IssueNewCookie = maybe_issue_new_cookie(
+                                    MochiReq, TimeLeft, Timeout),
                                 Req#httpd{user_ctx=#user_ctx{
                                     name=?l2b(User),
                                     roles=couch_util:get_value(<<"roles">>, UserProps, [])
-                                }, auth={FullSecret, TimeLeft < Timeout*0.9}};
+                                }, auth={FullSecret, IssueNewCookie}};
                             _Else ->
                                 Req
                         end;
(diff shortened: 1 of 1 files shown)


src/couch_httpd_auth.erl
```diff
@@ -240,13 +242,21 @@ cookie_authentication_handler(#httpd{mochi_req=MochiReq}=Req, AuthModule) ->
         end
     end.
 
+cookie_auth_should_issue_new_cookie(MochiReq, TimeLeft, Timeout) ->
```
kxepal: This is good name for test function, but not like this one. `maybe_issue_new_cookie` ?

adrienverge: You're right!

adrienverge: Thanks for you suggestion. I've updated the PR.

src/couch_httpd_auth.erl
```diff
@@ -240,13 +242,21 @@ cookie_authentication_handler(#httpd{mochi_req=MochiReq}=Req, AuthModule) ->
         end
     end.
 
+maybe_issue_new_cookie(MochiReq, TimeLeft, Timeout) ->
```
rnewson: should match on path_parts like elsewhere.

rnewson:
not sure on the patch itself but please no one merge before 2.0.

adrienverge:
Hey, thanks for reviewing.

> not sure on the patch itself

Do you mean not sure about the need for programmatical cookie renewal, or about the implementation? If there's a smarter/cleaner way of implementing this, I'll be glad to try!

rnewson:
I think it's already a problem that a cookie can be extended indefinitely (sure, it's invalidated by a password change but I don't think everyone realises that they only expire through inactivity).

The patch itself is close to 'done', see my previous comments.

tbh I would rather not merge this, and also remove the 10% renewal thing as well; forcing users to re-authenticate at least once per couch_httpd_auth timeout period. That's a larger conversation, and a controversial one, and definitely a post 2.0 thing, but I think many administrators expect cookies to expire after that timeout regardless of activity.

I don't want to discourage you from contributing, you've done a great job in your first patch. Authentication is a gnarly corner and CouchDB's auth system is fiddly in its own right.

A way forward, post 2.0, is a conversation about sessions and particularly about permanent sessions. We might support new authentication methods, like token authentication, to deal with this situation.

kxepal:
@adrienverge 
There will be more problem: authed user may eventually request `GET /_session` for the information about himself: username, roles etc. According the code, each time he'll do that, the new cookie will get issued. Sounds like not a good plan.

Isolating that behavior just for a some single method (like `POST`) could be good idea, but what @rnewson said, there need some discussion as the auth is very important thing and doesn't forgives any mistakes.

PR #197, opened 2016-09-12
Fix Windows version string determination

wohali:
For some reason, the COUCHDB_VERSION string on Windows is getting
suffixed whitespace. We fix this by string:strip(Version0, right) on the
variable before use.


rnewson:
+1

PR #198, opened 2016-09-13
Make sure authentication_db for clustered is considered as users db

iilyak:
Fixes the case when the following are configured to something other than _users
 - chttpd_auth.authentication_db
 - couch_httpd_auth.authentication_db

Introduce "couchdb.users_db_suffix" config key

Remove distinction between clustered and local users databases
configuration. We remove following configuration options and replace
them with "couchdb.users_db_suffix":

- chttpd_auth.authentication_db
- couch_httpd_auth.authentication_db

COUCHDB-2731

### DIFF
--- a/src/couch_server.erl
+++ b/src/couch_server.erl
@@ -119,9 +119,11 @@ maybe_add_sys_db_callbacks(DbName, Options) when is_binary(DbName) ->
 maybe_add_sys_db_callbacks(DbName, Options) ->
     DbsDbName = config:get("mem3", "shards_db", "_dbs"),
     NodesDbName = config:get("mem3", "nodes_db", "_nodes"),
-    IsReplicatorDb = path_ends_with(DbName, <<"_replicator">>),
-    IsUsersDb = DbName ==config:get("couch_httpd_auth", "authentication_db", "_users") orelse
-	path_ends_with(DbName, <<"_users">>),
+
+    IsReplicatorDb = path_ends_with(DbName, "_replicator"),
+    UsersDbSuffix = config:get("couchdb", "users_db_suffix", "_users"),
+    IsUsersDb = path_ends_with(DbName, "_users")
+        orelse path_ends_with(DbName, UsersDbSuffix),
     if
 	DbName == DbsDbName ->
 	    [sys_db | Options];
@@ -139,8 +141,10 @@ maybe_add_sys_db_callbacks(DbName, Options) ->
 	    Options
     end.
 
-path_ends_with(Path, Suffix) ->
-    Suffix == couch_db:normalize_dbname(Path).
+path_ends_with(Path, Suffix) when is_binary(Suffix) ->
+    Suffix =:= couch_db:dbname_suffix(Path);
+path_ends_with(Path, Suffix) when is_list(Suffix) ->
+    path_ends_with(Path, ?l2b(Suffix)).
 
 check_dbname(#server{}, DbName) ->
     couch_db:validate_dbname(DbName).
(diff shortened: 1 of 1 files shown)


davisp:
It seems a bit odd to have the system database feature flagged on an HTTP layer config option. Have we considered just removing both the chttpd and couch_httpd options and moving it to the `[couchdb]` section?

iilyak:
@davisp: We didn't consider it. Although in this case we would need to update documentation as well.
Do you mean something like follows?

```
[couchdb]

clustered_users_db_suffix = _users
node_local_users_db_suffix = _users
```

davisp:
Yes to in the [couchdb] section, no to the distinction between clustered and local. Just make "users_db_suffix = _users" and apply it the same across both. Not only do most people probably never change that (and honestly why is this even configurable to begin with?), who would want to have two different names for local vs clustered? That doesn't make any sense to me given that clients would have to have knowledge about that sort of thing and I've never once seen a client that even does anything other than the default _users db.

davisp:
Also, yes, we'd have to update the docs, but I was hoping that'd be trivial and just require moving a paragraph or two and tweaking for the name.

iilyak:
> @davisp:  who would want to have two different names for local vs clustered? 

Cloudant!

davisp:
Erm, what do we call it other than _users? And if we do, why do we do that?

iilyak:
`chttpd_auth.autentication_db = "<cluster>/users"`
`couch_httpd_auth.authentication_db = "users"`

davisp:
Isn't that obviously broken then? How would you ever match something with '/' in it to filename:basename(Anything)?

Is there an intent to make this different than _replicator databases where we just look at the last component of the path? couch_replicator_manager:is_replicator/1 actually even hard codes that check to be `<<"_replicator">> == couch_db:dbname_suffix(DbName).`

davisp:
Also it occurs to me that the check for the replicator here is broken since we hardcode that in couch_replicator_manager. We could be adding sys_db hooks to something the manager isn't looking at and not adding hooks for things it thinks are _replicator databases.

iilyak:
@davisp: How about the following?

```
[couchdb]
sys_dbs = _users, _replicator, _dbs, _nodes
```
I don't really like it since it gives the user ability to shout themselves in the foot. However it would simplify `maybe_add_sys_db_callbacks`.

iilyak:
No it is not going to work because we set different set of callbacks for _users and _replicator.

iilyak:
@davisp: I did fix `_replicator` part of a PR.
> Isn't that obviously broken then? How would you ever match something with '/' in it to filename:basename(Anything)?

basename is needed to extract last component of the path. The `couch_db:normalize_dbname` makes sure we strip `couch` extension, creation_time and whatnot.  
```
couch_db:1616: dbname_suffix_test_ ("long/co$mplex-/path+/_something" -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ (<<"long/co$mplex-/path+/_something">> -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ ("shards/00000000-3fffffff/long/co$mplex-/path+/_something.1415960794" -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ (<<"shards/00000000-3fffffff/long/co$mplex-/path+/_something.1415960794">> -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ ("shards/00000000-3fffffff/long/co$mplex-/path+/_something.1415960794.couch" -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ (<<"shards/00000000-3fffffff/long/co$mplex-/path+/_something.1415960794.couch">> -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ ("_something" -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ (<<"_something">> -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ ("shards/00000000-3fffffff/_something.1415960794" -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ (<<"shards/00000000-3fffffff/_something.1415960794">> -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ ("shards/00000000-3fffffff/_something.1415960794.couch" -> <<"_something">>)...ok
couch_db:1616: dbname_suffix_test_ (<<"shards/00000000-3fffffff/_something.1415960794.couch">> -> <<"_something">>)...ok
```

iilyak:
The PR has been re-written to merge two distinct settings `chttpd_auth.authentication_db` and `couch_httpd_auth.authentication_db` with a single one `couchdb.users_db_suffix`.

PR #199, opened 2016-09-16
Define couch version

iilyak:
This changes the way we set the value for `couch_server:get_version`. The `get_version` function is used in:
- version key of welcome message
- Server header returned in every response

This implementation defines a COUCH_VERSION macro. This is done in order to ensure that we keep using `{vsn, git},` in `src/couch.app.src`. There is also additional benefit of using the macro. It saves us an access to ets table. ETS is used for `application:get_key`. Since get_version is called for every request the saving should be significant. This would improve on previous analysis and work done by @robertkowalski [here](https://github.com/robertkowalski/kowalski.gd/blob/master/_posts/2016-02-10-high-performance-erlang-finding-bottlenecks-couchdb-1.md).

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,2 +1,0 @@
-{erl_opts, [{platform_define, "win32", 'WINDOWS'}]}.
-{eunit_compile_opts, [{platform_define, "win32", 'WINDOWS'}]}.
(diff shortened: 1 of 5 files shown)


kxepal:
+1

rnewson:
+1, nice and clean.

PR #201, opened 2016-10-04
Add max_document_size checking for multipart PUT requests

nickva:
Previously multipart/related PUT requests didn't check maximum request sizes.

This commit checks content-length and compares that with the maximum.

This means keeping the current "semantics" of max_document_size which actually
means "max request size". But this makes the check more efficient and can
be done earlier in request processing time.

Jira: COUCHDB-3174


davisp:
+1 but we'll want to document the new behavior somewhere. Also, you don't need the Jira: in front of your COUCHDB-3174 in the commit message.

PR #202, opened 2016-10-04
Fix CouchJS character replacement

davisp:
This was a bad backport from an old bug. We accidentally backed up when
looking at the second half of a surrogate pair. Instead the backup
should only happen when we see a low half of a surrogate pair with no
preceding high half.

COUCHDB-3173


rnewson:
+1

PR #206, opened 2016-10-13
handle errors terms from fabric

iilyak:
Test suite for https://github.com/apache/couchdb-chttpd/pull/146


iilyak:
The PR has been rebased

jaydoane:
LGTM, with the caveat that we seem to be unnecessarily running the same test twice.

PR #208, opened 2016-10-27
format process state

iilyak:
Companion PR for https://github.com/apache/couchdb-couch-replicator/pull/51

### DIFF
--- a/src/couch_db.erl
+++ b/src/couch_db.erl
@@ -36,6 +36,7 @@
 -export([normalize_dbname/1]).
 -export([validate_dbname/1]).
 -export([dbname_suffix/1]).
+-export([format_db_record/1]).
 
 -include_lib("couch/include/couch_db.hrl").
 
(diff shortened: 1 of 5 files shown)


src/couch_util.erl
```diff
@@ -584,3 +585,14 @@ with_proc(M, F, A, Timeout) ->
         erlang:demonitor(Ref, [flush]),
         {error, timeout}
     end.
+
+
+take_kv_values(List, Keys) ->
+    lists:foldr(fun(Key, {Acc, KVList}) ->
```
tonysun83: derp, answered my own question: A body-recursive list function and a tail-recursive function that calls lists:reverse/1 at the end will use the same amount of memory.

src/couch_db.erl
```diff
@@ -1560,6 +1561,30 @@ is_systemdb(DbName) when is_list(DbName) ->
 is_systemdb(DbName) when is_binary(DbName) ->
     lists:member(dbname_suffix(DbName), ?SYSTEM_DATABASES).
 
+format_db_record(#db{} = Db) ->
```
tonysun83: should we make this more flexible by adding format_db_record/2 that takes in a separate list rather and then having it to default to this list when users don't provide the db keys they want? Just thinking longer term.

iilyak:
stale PR

PR #209, opened 2016-10-27
remove compact files on failure

iilyak:
In the case when `compact.meta` file is corrupted the compaction fails over and over. 
This fixes the cyclic failure aspect by use of a `.compact.crash` flag file.
The logic is:

``` python
if there is an exception and  `.compact.crash` exists:
     remove(.compact.{meta|data})
```

COUCHDB-3216

### DIFF
--- a/src/couch_db_updater.erl
+++ b/src/couch_db_updater.erl
@@ -18,6 +18,7 @@
 -export([btree_by_seq_split/1, btree_by_seq_join/2, btree_by_seq_reduce/2]).
 -export([make_doc_summary/2]).
 -export([init/1,terminate/2,handle_call/3,handle_cast/2,code_change/3,handle_info/2]).
+-export([delete_compact_files/2]).
 
 -include_lib("couch/include/couch_db.hrl").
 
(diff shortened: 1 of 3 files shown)


src/couch_db_updater.erl
```diff
@@ -1454,3 +1450,13 @@ default_security_object(_DbName) ->
...
+
+delete_compact_files(RootDir, FullFilepath) ->
+    lists:foreach(fun(Ext) ->
+        couch_file:delete(RootDir, FullFilepath ++ Ext)
+    end, [".compact", ".compact.data", ".compact.meta"]).
+
+purge_compact_files(FullFilepath) ->
+    lists:foreach(fun(Ext) ->
+        file:delete(FullFilepath ++ Ext)
+    end, [".compact", ".compact.data", ".compact.meta"]).
```
jaydoane: Since you repeat this list, any reason to not create a 
`define(COMPACT_SUFFIXES, [".compact", ".compact.data", ".compact.meta"]).`
?

rnewson:
Can you say more on how .compact.meta files are being corrupted? That's quite serious and needs addressing first.

iilyak:
@rnewson: They are corrupted from couchdb perspective when we do upgrade from Cloudant version to apache version of couchdb. Even though this problem only affects Cloudant this commit improves resilience of the compactor against failures.

jaydoane:
+1

rnewson:
I am a little nervous here. The explanation makes sense though. 

Before merging, more +1's please.

iilyak:
closing this due to inactivity

PR #210, opened 2016-11-03
Include file_path on errors from couch_file

iilyak:
Add PathFile as part of a reason for following errors:
 - read_beyond_eof
 - exceed_pread_limit exceptions

### DIFF
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -428,10 +428,10 @@ handle_call({pread_iolist, Pos}, _From, File) ->
         % up to 8Kbs of read ahead
         read_raw_iolist_int(File, Pos, ?READ_AHEAD - (Pos rem ?SIZE_BLOCK))
     catch
-    throw:read_beyond_eof ->
-        throw(read_beyond_eof);
-    throw:{exceed_pread_limit, Limit} ->
-        throw({exceed_pread_limit, Limit});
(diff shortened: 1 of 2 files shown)


src/couch_file.erl
```diff
@@ -625,10 +627,12 @@ read_raw_iolist_int(#file{fd = Fd, pread_limit = Limit} = F, Pos, Len) ->
     case Pos + TotalBytes of
     Size when Size > F#file.eof + ?READ_AHEAD ->
         couch_stats:increment_counter([pread, exceed_eof]),
-        throw(read_beyond_eof);
+        {_Fd, Filepath} = get(couch_file_fd),
+        throw({read_beyond_eof, Filepath});
```
eiri: This is wrong, isn't it? You are throwing `{read_beyond_eof, Filepath}` here, but then trying to catch `throw:read_beyond_eof` above, in `handle_call`.

iilyak: @eiri: Should be fixed now

eiri: Yes and I think it's better this way.

Theoretically moving `get(couch_file_fd)` out of the `case` would allow not to call it in each clause, making it DRY'er and all of that, but I don't feel too strongly about it, feel free to keep it as it is.

eiri:
Not sure why travis is stuck here, but locally tests are passing. +1

PR #211, opened 2016-11-11
Allow fixing users' documents

iilyak:
Allow adding missing fields:
- type
- roles

Allow update of the type field from incorrect value to the correct one
(currently it suppose to have value 'user').

COUCHDB-3231


rnewson:
+1

PR #214, opened 2016-11-23
Handle open_result message that arrives after the delete

rnewson:
Opening a database is asynchronous, therefore it's possible for the
database to have been deleted before the open is complete. The
open_result code assumes this doesn't happen, so when it does,
couch_server crashes, taking the couch_dbs ETS table with it.

Explicitly handle the case and keep on keeping on.

COUCHDB-3241


eiri:
+1

PR #215, opened 2016-11-28
Implement couch_file:format_status to log filepath

iilyak:
Erlang OTP logs the fact when gen_server behavior crashes. However in case of `couch_file` the filename is not part of the state. So it is quite hard to figure out what happened given the current log entry format. This PR adds information about filepath to the log entry. Keep in mind though that the filepath value is captured on couch_file:start_link. It is not representing the current file name if file is renamed or moved.


rnewson:
+1

iilyak:
Closing it in favor of https://github.com/apache/couchdb/pull/1601 (merged already).

PR #216, opened 2016-11-30
Make couchjs -S option take effect

nickva:
Previously it was used to set JS context's stack chunk size.

Instead, to be effective it should set max GC size of the runtime.

Stack chunk size was set to the recommended value: 8K

Reference:

https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_reference/JS_NewContext

COUCHDB-3245

### DIFF
--- a/priv/couch_js/main.c
+++ b/priv/couch_js/main.c
@@ -387,11 +387,11 @@ main(int argc, const char* argv[])
 
     couch_args* args = couch_parse_args(argc, argv);
 
-    rt = JS_NewRuntime(64L * 1024L * 1024L);
+    rt = JS_NewRuntime(args->stack_size);
     if(rt == NULL)
         return 1;
 
-    cx = JS_NewContext(rt, args->stack_size);
+    cx = JS_NewContext(rt, 8L * 1024L);
     if(cx == NULL)
         return 1;
 


jaydoane:
+1

Though it's curious how the previous fix https://issues.apache.org/jira/browse/COUCHDB-1792 was apparently never applied. However, it clearly seems to be the right thing to do.

nickva:
Thanks @jaydoane 

I think the fix was applied then switched back during a refactoring? 

It is interesting why it works though. I think it probably masking a fundamental and tricky bug with garbage collection. By essentially stopping it from running.

However for practical reasons we saw that at least this brings us back to behavior that we've had before in production.

jaydoane:
Yes, maybe reverted during a refactor, since all the files that patch applied to (sm170.c, sm180.c, sm185.c) no longer exist. I guess that's easy to do with an absence of unit tests ;)

davisp:
+1. But add a reference to Randall's commit that somehow got reverted.

davisp:
This one: https://github.com/apache/couchdb-couch/commit/62dafe81e13d7f8e27e95057e65f76d534aa2313

And make sure the rest of that commit is applied as well.

PR #218, opened 2016-12-13
Make revision generation deterministic

davisp:
This removes the influence of the attachment disk information when
generating revisions when a document is being recreated (ie, it existed
once but has been deleted).

COUCHDB-3255

### DIFF
--- a/src/couch_db.erl
+++ b/src/couch_db.erl
@@ -860,7 +860,7 @@ prep_and_validate_replicated_updates(Db, [Bucket|RestBuckets], [OldInfo|RestOldI
 
 
 
-new_revid(#doc{body=Body, revs={OldStart,OldRevs}, atts=Atts, deleted=Deleted}) ->
+new_revid(#doc{body=Body0, revs={OldStart,OldRevs}, atts=Atts, deleted=Deleted}) ->
     DigestedAtts = lists:foldl(fun(Att, Acc) ->
         [N, T, M] = couch_att:fetch([name, type, md5], Att),
         case M == <<>> of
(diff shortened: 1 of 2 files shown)


davisp:
For color, the reason we ran into this now is because in a cluster we rely on the generated revisions to be exactly the same. We happened to have a user that was basically running "DELETE /dbname/foo", "PUT /dbname/foo" with an attachment a few times a day. Basically, unless you do that specific pattern on a 2.0 cluster you'll never see an issue.

@tonysun is gonna write a test to add to this PR tonight so we'll have that going forward.

tonysun83:
@davisp: added the test and changed the typo. One thing I'm a little fuzzy over is the fact that we're extracting the attachment out of the calculation and only using the body to calculate the new revid. If that's the case, then it doesn't matter what the contents of the attachment is right? As long as the body is the same, the rev should be the same? In my test, I left the attachment content as the same, and we get the same rev id. However, if I change the content, then the rev id is different in a different/newly created database. Is that expected?

davisp:
The name, content type, and body (via md5) determine the revision. If you specify a different body then you should expect a different revision id, yes.

davisp:
Oh, it looks like maybe that the doc id isn't part of the revision calculation was confusing. I suddenly can't remember why it doesn't but it doesn't really make a difference cause you wouldn't be mixing revisions from different doc ids anyway.

davisp:
I cleaned up the test a bit and squashed everything.

Also, for posterity, I checked that this test failed without the change to new_revid.

iilyak:
+1. Tests pass locally. 

```
Compiled test/chttpd_endpoints_tests.erl
    Running test function(s):
      couch_db_mpr_tests:couch_db_mpr_test_/0
======================== EUnit ========================
multi-part attachment tests
  couch_db_mpr_tests: couch_db_mpr_test_...[0.256 s] ok
  [done in 0.260 s]
=======================================================
  Test passed.
```

tonysun83:
+1, ran the repro script as well to confirm correct behavior

PR #219, opened 2016-12-21
Don't trap exits in couch_file

rnewson:
It turns out to be impossible to trap the exit message from the parent
process in a gen_server, which is the only reason couch_file traps
exits in the first place (i.e, this has never worked).

This commit also changes the hot upgrade code to handle this
transition (but not the previous one).

COUCHDB-3259

### DIFF
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -21,7 +21,7 @@
 -define(MONITOR_CHECK, 10000).
 -define(SIZE_BLOCK, 16#1000). % 4 KiB
 -define(READ_AHEAD, 2 * ?SIZE_BLOCK).
--define(IS_OLD_STATE(S), tuple_size(S) /= tuple_size(#file{})).
+-define(IS_OLD_STATE(S), is_pid(S#file.db_monitor)).
 -define(PREFIX_SIZE, 5).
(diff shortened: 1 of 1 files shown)


nickva:
It should be possible to trap exit messages. Message is delivered to `terminate` callback instead of `handle_info`.

Notice line: https://github.com/erlang/otp/blob/maint/lib/stdlib/src/gen_server.erl#L389

rnewson:
That we're in terminate means we didn't trap it. :)

nickva:
touché

rnewson:
It's specifically that a process can't trap the exit of its parent process in a gen_server, which is the very code you pointed at. This wasn't realised at the time couch_file was first written.

nickva:
Right, right. I initially thought it was about doing something before exiting anyway then handling thing in terminate might have worked, but if we continue working terminate handling won't work

davisp:
+1

davisp:
A note on the earlier conversation, the point of the original 'EXIT' handling code was so that couch_file could outlive its parent. But that's not possible because even if we trap exits, gen_server forces an exit if the signal is from the parent. However, since all we were doing was ignoring a `normal` exit and closing on anything else, that's the same exact behavior we'd get if we just didn't trap exit signals.

PR #221, opened 2017-01-18
Improve errors on invalid "count" parameter in /_uuids

eiri:
Added validation on "count" parameter to be positive and changed return code for "count" exceeding maximum from "Forbidden" to "Bad Request"

COUCHDB-3273

### DIFF
--- a/src/couch_httpd_misc_handlers.erl
+++ b/src/couch_httpd_misc_handlers.erl
@@ -129,7 +129,7 @@ handle_uuids_req(#httpd{method='GET'}=Req) ->
     Max = list_to_integer(config:get("uuids","max_count","1000")),
     Count = try list_to_integer(couch_httpd:qs_value(Req, "count", "1")) of
         N when N > Max ->
-            throw({forbidden, <<"count parameter too large">>});
+            throw({bad_request, <<"count parameter too large">>});
         N when N < 0 ->
             throw({bad_request, <<"count must be a positive integer">>});
         N -> N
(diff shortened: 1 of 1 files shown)


kxepal:
I think we need some tests on this, because that's not a first time we fixing handling this parameter right.

But in anyway, +1.

src/couch_httpd_misc_handlers.erl
```diff
@@ -129,7 +129,9 @@ handle_uuids_req(#httpd{method='GET'}=Req) ->
     Max = list_to_integer(config:get("uuids","max_count","1000")),
     Count = try list_to_integer(couch_httpd:qs_value(Req, "count", "1")) of
         N when N > Max ->
-            throw({forbidden, <<"count parameter too large">>});
+            throw({bad_request, <<"count parameter too large">>});
+        N when N < 0 ->
+            throw({bad_request, <<"count must be a positive integer">>});
```
kxepal: May be also edit line 138 to note that we want positive integer, not just an integer?

eiri:
@kxepal I agree on tests, but it's a bit awkward, with handler been here and tests need to be in `chttpd`, so they guarantee to fail. I'll add them after this get merged and added to rebar's config, so CI will stay green.

kxepal:
@eiri ah, indeed, I forgot that testing is quite hard for now /: Ok, then.

rnewson:
@wohali noticed that the test associated with this, that confirms we send a 403 for the forbidden case has been broken since this change (several months!)

When changing http semantics, please update tests and documentation to match.

cc @eiri @kxepal.

rnewson:
(in the couchdb javascript test suite)

kxepal:
@rnewson Good reminder. Thanks! Sorry that I missed that moment /:

PR #222, opened 2017-01-19
Reset EOF if a partial write was possible

rnewson:
We can't know if one or more bytes were written by a file:write/2 call
that results in an error and so it is not correct to leave #file.eof
at its original value. In the event of error, use file:position(Fd,
eof) to find the new, true length of the file, and update #file{}
accordingly.

COUCHDB-3274

### DIFF
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -476,7 +476,7 @@ handle_call({append_bin, Bin}, _From, #file{fd = Fd, eof = Pos} = File) ->
     ok ->
         {reply, {ok, Pos, Size}, File#file{eof = Pos + Size}};
     Error ->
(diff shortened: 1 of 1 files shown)


davisp:
+1

Also a note that this was caught by inspection and remembering we've seen weirdness when a hdd runs out of space. A possible cause would be getting enospc errors from writes and then using an incorrect eof after some space had been recovered.

rnewson:
I noted that in the JIRA ticket but worth repeating here.

I struggled (and ultimately failed) to add a test here, as it involves either a full mocking out of `file` or setting up a very small device (e.g, a 1mb ram disk). I did the ram disk approach for manual testing, fwiw.

PR #223, opened 2017-01-27
Remove 8kB read-ahead from couch_file

nickva:
In production it showed an increased input Erlang IO and increased binary
memory usage.

See attached file in Jira ticket:

COUCHDB-3284

### DIFF
--- a/src/couch_file.erl
+++ b/src/couch_file.erl
@@ -20,7 +20,6 @@
 -define(INITIAL_WAIT, 60000).
 -define(MONITOR_CHECK, 10000).
 -define(SIZE_BLOCK, 16#1000). % 4 KiB
--define(READ_AHEAD, 2 * ?SIZE_BLOCK).
 -define(IS_OLD_STATE(S), is_pid(S#file.db_monitor)).
 -define(PREFIX_SIZE, 5).
 -define(DEFAULT_READ_COUNT, 1024).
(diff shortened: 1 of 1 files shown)


kxepal:
@nickva 
You're going to revert [this optimization](https://github.com/apache/couchdb-couch/commit/c7b30071ac1a4166d44fc10f760a1a8ae2a8ae5c). Won't this impact on something else?

kxepal:
Hm...just checked your graphs in JIRA. They looks very nice.

nickva:
Out of curiosity create a branch which allows configuring file:open's read_ahead parameter:

https://github.com/cloudant/couchdb-couch/commit/3e35ae40f71fa84a3453b178caf774147300e87d

davisp:
Awesome find. I've definitely been perplexed by the I/O changes since starting to deploy Couch 2.0 code. This definitely seems to explain both the io differences and the increase in memory usage. 

The original commit even calls this a micro optimization so I'm inclined to completely remove it rather than make it optional. @kxepal, are you wanting to see a config option or are the graphs on the ticket convincing enough to just drop it?

Huge +1 from me.

kxepal:
@davisp 
No, I'm waiting to have this merged (: 
I wrote first post without knowledge about graphs, which are impressive - that's why reverting optimization for another optimization I found a bit strange.

+1 - that's what I forgot to add.

rnewson:
+1

rnewson:
for clarity, we strongly believe the readahead to be a significant performance _regression_. It's not an improvement in any scenario we can see.

PR #224, opened 2017-01-30
Don't crash on unexpected validation's error type

eiri:
Validation function can throw JSON with error type different than forbidden or unauthorized.

This patch makes couch return unknown_error instead of crashing query server.

COUCHDB-3286


kxepal:
+1

nickva:
+1 as well

Was curious how it worked. 

```
In [33]: db = srv.create('db')
In [34]: db['_design/des'] = {'validate_doc_update':'''function(nd,od,ctx) { throw({zig:"blah"}); }'''}
In [35]: db['x'] = {}
ServerError: (500, (u'unknown_error', u'function_clause'))
```

with fix

```
ServerError: (500, (u'unknown_error', u'blah'))
```

PR #225, opened 2017-02-01
Remove dead code from couch_file

nickva:
This code was left over after removing 8kB read-ahead

https://github.com/apache/couchdb-couch/pull/223/commits/d52a5335d930d11ade4953c8576d22f55872ff6f

COUCHDB-3284


iilyak:
+1

PR #227, opened 2017-02-08
Add test_request:delete/3

iilyak:
Sometimes it is useful to be able to pass headers in test case to a delete request. 
Adding delete/3 to be able to do so.

### DIFF
--- a/src/test_request.erl
+++ b/src/test_request.erl
@@ -20,45 +20,46 @@
 -export([request/3, request/4]).
 
 get(Url) ->
-    request(get, Url, []).
+    get(Url, []).
 
 get(Url, Headers) ->
-    request(get, Url, Headers).
+    get(Url, Headers, []).
+
 get(Url, Headers, Opts) ->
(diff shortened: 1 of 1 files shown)


src/test_request.erl
```diff
@@ -51,6 +51,9 @@ delete(Url) ->
 delete(Url, Opts) ->
     request(delete, Url, [], [], Opts).
```
eiri: Maybe make this line to use this new function?

iilyak: @eiri: I used the same approach as in other functions get/post/put. We just call request from there. Personally I prefer the approach you suggest. Should I refactor all of them for consistency?

eiri: fair point. I don't know, tbh, I don't feel particularly strong about it, but using smaller arity functions as defaults for same functions with bigger arity is pretty standard erlang idiom, so if you up to a bigger change I'll +1 it.

iilyak:
@eiri: done

PR #228, opened 2017-02-11
Couchdb 3298 improve couch btree chunkify

davisp:
This PR adds two slight tweaks to the couch_btree:chunkify/1 function.

First, rather than create a larger number of evenly sized chunks it will create as few chunks under the configurable chunk threshold as possible with the final chunk possibly being less full than average.

Second, it prevents the return of a final chunk that has a single key. In some pathological cases we could end up with a branch of the tree that had multiple levels of nodes that had a single key. This is caused when a reduce function is returning values that are larger than the chunk threshold.

### DIFF
--- a/src/couch_btree.erl
+++ b/src/couch_btree.erl
@@ -342,11 +342,9 @@ complete_root(Bt, KPs) ->
 % it's probably really inefficient.
 
 chunkify(InList) ->
-    BaseChunkSize = get_chunk_size(),
+    ChunkThreshold = get_chunk_size(),
     case ?term_size(InList) of
-    Size when Size > BaseChunkSize ->
-        NumberOfChunksLikely = ((Size div BaseChunkSize) + 1),
-        ChunkThreshold = Size div NumberOfChunksLikely,
+    Size when Size > ChunkThreshold ->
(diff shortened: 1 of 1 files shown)


iilyak:
Tested manually with the following:
```
LL = couch_btree:chunkify(
   [{list_to_binary(couch_util:to_hex(crypto:rand_bytes(10))), binary:copy(<<"x">>, 1300)}
       || _ <- lists:seq(1, 10)]).
length(LL).

 Generator = fun(Size, N) ->
   [{list_to_binary(couch_util:to_hex(crypto:rand_bytes(10))), binary:copy(<<"x">>, Size)}
      || _ <- lists:seq(1, N)]
 end.

 LL = couch_btree:chunkify(Generator(1300, 3)).
 [Last | _] = lists:reverse(LL).
 [{"Number of chunks", length(LL)}, {"Elements in last chunk", length(Last)}].
```

The output without this change is:
```
[
  {"Number of chunks",5},
  {"Elements in last chunk",2}
]
```

The output with this change is:
```
[
  {"Number of chunks",1},
  {"Elements in last chunk",3}
]
```

+1.

PR #229, opened 2017-02-21
Return error row instead of crashing

tonysun83:
When input is invalid for built in reducers, we return an error row
instead of crashing.  Thanks to @davisp for providing the elegant
solution.

The results now look like this:
```
{
    "rows": [
        {
            "key": null,
            "value": {
                "key": [],
                "value": [
                    {
                        "a": 1
                    },
                    {
                        "a": 2
                    },
                    {
                        "a": 3
                    },
                    {
                        "a": 4
                    }
                ],
                "builtin_reduce_error": "The _sum function requires that map values be numbers, arrays of numbers, or objects, not '[{[{<<\"b\">>,\n                                                                                               1}]},\n                                                                                            {[{<<\"b\">>,\n                                                                                               2}]},\n                                                                                            {[{<<\"b\">>,\n                                                                                               3}]},\n                                                                                            {[{<<\"b\">>,\n                                                                                               4}]}]'. Objects cannot be mixed with other data structures. Objects can be arbitrarily nested, provided that the values for all fields are themselves numbers, arrays of numbers, or objects."
            }
        }
    ]
}
```
I didn't change the initial error message, as it was quite informative. Another bonus for this change
is that the log isn't flooded with error messages.

COUCHDB-3305

### DIFF
--- a/src/couch_query_servers.erl
+++ b/src/couch_query_servers.erl
@@ -27,7 +27,7 @@
 -include_lib("couch/include/couch_db.hrl").
 
 -define(SUMERROR, <<"The _sum function requires that map values be numbers, "
-    "arrays of numbers, or objects, not '~p'. Objects cannot be mixed with other "
+    "arrays of numbers, or objects. Objects cannot be mixed with other "
     "data structures. Objects can be arbitrarily nested, provided that the values "
     "for all fields are themselves numbers, arrays of numbers, or objects.">>).
 
@@ -142,25 +142,34 @@ os_rereduce(Lang, OsRedSrcs, KVs) ->
 builtin_reduce(_Re, [], _KVs, Acc) ->
     {ok, lists:reverse(Acc)};
 builtin_reduce(Re, [<<"_sum",_/binary>>|BuiltinReds], KVs, Acc) ->
-    Sum = builtin_sum_rows(KVs),
+    Sum = builtin_sum_rows(KVs, []),
     builtin_reduce(Re, BuiltinReds, KVs, [Sum|Acc]);
 builtin_reduce(reduce, [<<"_count",_/binary>>|BuiltinReds], KVs, Acc) ->
     Count = length(KVs),
     builtin_reduce(reduce, BuiltinReds, KVs, [Count|Acc]);
 builtin_reduce(rereduce, [<<"_count",_/binary>>|BuiltinReds], KVs, Acc) ->
-    Count = builtin_sum_rows(KVs),
+    Count = builtin_sum_rows(KVs, []),
     builtin_reduce(rereduce, BuiltinReds, KVs, [Count|Acc]);
 builtin_reduce(Re, [<<"_stats",_/binary>>|BuiltinReds], KVs, Acc) ->
     Stats = builtin_stats(Re, KVs),
     builtin_reduce(Re, BuiltinReds, KVs, [Stats|Acc]).
 
-builtin_sum_rows(KVs) ->
-    lists:foldl(fun([_Key, Value], Acc) -> sum_values(Value, Acc) end, 0, KVs).
 
-sum_values({Props}, 0) ->
-    {Props};
-sum_values({Props}, {AccProps}) ->
-    {sum_objects(lists:sort(Props), lists:sort(AccProps))};
+builtin_sum_rows([], Acc) ->
+    Acc;
+builtin_sum_rows([[_Key, Value] | RestKVs], Acc) ->
+    try sum_values(Value, Acc) of
+        NewAcc ->
+            builtin_sum_rows(RestKVs, NewAcc)
+    catch
+        throw:{builtin_reduce_error, Obj} ->
+            Obj;
+        throw:{invalid_value, Reason, Cause} ->
+            {[{<<"error">>, <<"builtin_reduce_error">>},
+                {<<"reason">>, Reason}, {<<"caused_by">>, Cause}]}
+    end.
+
+
 sum_values(Value, Acc) when is_number(Value), is_number(Acc) ->
     Acc + Value;
 sum_values(Value, Acc) when is_list(Value), is_list(Acc) ->
@@ -169,6 +178,19 @@ sum_values(Value, Acc) when is_number(Value), is_list(Acc) ->
     sum_arrays(Acc, [Value]);
 sum_values(Value, Acc) when is_list(Value), is_number(Acc) ->
     sum_arrays([Acc], Value);
+sum_values({Props}, Acc) ->
+    case lists:keyfind(<<"error">>, 1, Props) of
+        {<<"error">>, <<"builtin_reduce_error">>} ->
+            throw({builtin_reduce_error, {Props}});
+        false ->
+            ok
+    end,
+    case Acc of
+        0 ->
+            {Props};
+        {AccProps} ->
+            {sum_objects(lists:sort(Props), lists:sort(AccProps))}
+    end;
 sum_values(Else, _Acc) ->
     throw_sum_error(Else).
 
@@ -490,7 +512,7 @@ ret_os_process(Proc) ->
     ok.
 
 throw_sum_error(Else) ->
-    throw({invalid_value, iolist_to_binary(io_lib:format(?SUMERROR, [Else]))}).
+    throw({invalid_value, ?SUMERROR, Else}).
 
 throw_stat_error(Else) ->
     throw({invalid_value, iolist_to_binary(io_lib:format(?STATERROR, [Else]))}).
@@ -498,6 +520,17 @@ throw_stat_error(Else) ->
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
+
+builtin_sum_rows_negative_test() ->
+    A = [{[{<<"a">>, 1}]}, {[{<<"a">>, 2}]}, {[{<<"a">>, 3}]}],
+    E = {[{<<"error">>, <<"builtin_reduce_error">>}]},
+    ?assertEqual(E, builtin_sum_rows([["K", E]], [])),
+    % The below case is where the value is invalid, but no error because
+    % it's only one document.
+    ?assertEqual(A, builtin_sum_rows([["K", A]], [])),
+    {Result} = builtin_sum_rows([["K", A]], [1, 2, 3]),
+    ?assertEqual({<<"error">>, <<"builtin_reduce_error">>},
+        lists:keyfind(<<"error">>, 1, Result)).
 
 sum_values_test() ->
     ?assertEqual(3, sum_values(1, 2)),
@@ -511,6 +544,19 @@ sum_values_test() ->
             {<<"f">>,1}, {<<"g">>,2}]},
     ?assertEqual(Z, sum_values(X, Y)),
     ?assertEqual(Z, sum_values(Y, X)).
+
+sum_values_negative_test() ->
+    % invalid value
+    A = [{[{<<"a">>, 1}]}, {[{<<"a">>, 2}]}, {[{<<"a">>, 3}]}],
+    B = ["error 1", "error 2"],
+    C = [<<"error 3">>, <<"error 4">>],
+    KV = {[{<<"error">>, <<"builtin_reduce_error">>},
+        {<<"reason">>, ?SUMERROR}, {<<"caused_by">>, <<"some cause">>}]},
+    ?assertThrow({invalid_value, _, _}, sum_values(A, [1, 2, 3])),
+    ?assertThrow({invalid_value, _, _}, sum_values(A, 0)),
+    ?assertThrow({invalid_value, _, _}, sum_values(B, [1, 2])),
+    ?assertThrow({invalid_value, _, _}, sum_values(C, [0])),
+    ?assertThrow({builtin_reduce_error, KV}, sum_values(KV, [0])).
 
 stat_values_test() ->
     ?assertEqual({1, 2, 0, 1, 1}, stat_values(1, 0)),


iilyak:
What about other kinds of errors? Should we consider returning errors as:
```
"value": {
                "key": [],
                "value": <something>,
                "errors": {
                       "builtin_reduce": [
                              "The _sum function requires that map values be numbers, arrays of numbers, or objects, not '[{[{<<\"b\">>,\n",
                              "(whatever other errors we might have)" 
                             ]
                }
}
```

kxepal:
How the clients should handle such kind of response? 
It has the result data and it has some errons. Should the result be handled? Or an exception being throw?

I think there should be two solid states: success aka "Here are your data" and error aka "Sorry, fix me ASAP" without any misinterpretations.

kxepal:
For map functions errors we return the response in the following format:
```
{"total_rows":14,"offset":0,"rows":[
{"key":1,"error":"not_found"}
]}
```

It's hard to mismatch it or misinterpreter - you have to handle these errors somehow or crash. You cannot ignore them. I suggest to follow the same strategy in this case as well.

tonysun83:
@iilyak : The only two error types I can think of are when the input value is invalid and also what @kxepal  mentioned (the map function errors) which I forgot to take care of. In any case, I think if there are multiple reduces, then there would be multiple error rows per reduce call

```
{
    "rows": [
        {
            "key": null,
            "value": {
                "key": [],
                "value": <some_value>
                "builtin_reduce_error":...."
        },
        {
            "key": null,
            "value": {
                "key": [],
                "value": <some_value>
                "builtin_reduce_error":...."
        },
    ]
}
```

What about the following response format:
```
{
    "rows": [
        {
            "key": null,
            "value": {
                "error":  {"builtin_reduce_error": { "key" : <some_key>,  "value": <some_value>}
        },
    ]
}
```
So if it is from a map function then it would be:
```
{
    "rows": [
        {
            "key": null,
            "value": {
                "error":  {"builtin_reduce_error": { "key" : 1,  "value": "not_found"}
        },
    ]
}
```

kxepal:
It's still easy to confuse with the regular response because technically I can produce the `{'error': ...}` value with a reduce function and that's wouldn't be an error case.

I suggest to do the same as what we do for map functions errors: remove the value key and add the error one at the same level instead. You'll have to crash if error happened or you'll be forced to process the error field correctly. 

So may be something like this?:
```
{
    "rows": [
        {
            "key": null,
            "error":  "builtin_reduce_error",  
            "reason":  "The _sum function requires that map values be numbers, arrays of numbers, or objects, not '[{[{<<\"b\">>,\n"
        },
    ]
}
```

I used the `error-reason` is a common pair of keys we throw in case of errors, so it's well known pattern to handle. The errors dict is a something new, but it could work as well if  we can accumulate all the errors and fail fast.

davisp:
@iilyak For other types of errors if/when we add them I would expect the error to change. We're not collecting errors, the user will just get the first one that's found when we rereduce our output up the tree.

@kxepal I like your suggestion but I'd change it slightly so that we don't give a key instead of the value. The reason for this is so that users see the value that caused the error that's being reported. Also the key is pretty much a lie anyway since it's not necessarily a key from the key/value and instead can be from a key/pointer node.

Though I also see the point around confusion of providing a value there. So I renamed it to "error_value". 

So I'd tweak it to something like:

```json
{
    "error_value": $value_that_caused_the_error,
    "error": "builtin_reduce_error",
    "reason": "Blahbity blah blah..."
}
```

I'd also then go ahead and reword those error messages to not include the value again since we'll have it in the response.

@tonysun83 For the comment about multiple reduces, that is incorrect. Even though you're seeing them in the code down there, the way those are exposed to users is through different views. And each reduce is independent. So while something my not be valid for _stats, it might be valid for _sum and/or a custom reduce function.

kxepal:
@davisp 
Good points. I like this. May be only reword `error_value` to  `cause` or `caused_by`?

tonysun83:
@kxepal : going through the code path now, but I think adding a new "error" part would require a change outside of couch_query_servers. Probably couch_mrview or chttpd?

tonysun83:
@davisp: thanks for the clarification about the multiple reduces

One of the problems I had with understanding the code was figuring out how to modify the response format at the higher level (i.e the new format above) with only modifying couch_query_servers. IIUC, this will require changes to possibly chttpd or couch_mrview?

tonysun83:
Ok, I think I figured it out. Just need to add a clause here with the return value from couch_query_servers:

https://github.com/apache/couchdb-fabric/blob/master/src/fabric_view.erl#L261-L270

tonysun83:
@kxepal : To get the format that we want above, I'm catching the throw from an invalid input, and then forcing couch_query_servers to return ```{builtin_reduce_error, Reason, Value}```.  Then, I bubble up that return value and finally modify the message here:
https://github.com/apache/couchdb-couch-mrview/blob/master/src/couch_mrview_http.erl#L431

with
```
row_to_json({builtin_reduce_error, Reason, Value}) ->
    Obj = {[{error, builtin_reduce_error}, {reason, Reason}, {caused_by, Value}]},
    ?JSON_ENCODE(Obj);
```
However, maybe I'm not approaching this correctly since I am still confused with: "You'll have to crash if error happened or you'll be forced to process the error field correctly." Can you clarify a bit?

Tomorrow, I'm going to test the scenario with a map function errors so I can better understand what's going on.

tonysun83:
This is what it looks like now:

```
{
    "rows": [
        {
            "error": "builtin_reduce_error",
            "reason": "The _sum function requires that map values be numbers, arrays of numbers, or objects, not. Objects cannot be mixed with other data structures. Objects can be arbitrarily nested, provided that the values for all fields are themselves numbers, arrays of numbers, or objects.",
            "cause_by": [
                {
                    "a": 1
                },
                {
                    "a": 2
                },
                {
                    "a": 3
                },
                {
                    "a": 4
                }
            ]
        }
    ]
}
```

Also added  https://github.com/apache/couchdb-couch-mrview/pull/66
and https://github.com/apache/couchdb-fabric/pull/90 to support this change.

Hopefully this is the right approach.

src/couch_query_servers.erl
```diff
@@ -27,7 +27,7 @@
 -include_lib("couch/include/couch_db.hrl").
 
 -define(SUMERROR, <<"The _sum function requires that map values be numbers, "
-    "arrays of numbers, or objects, not '~p'. Objects cannot be mixed with other "
+    "arrays of numbers, or objects, not. Objects cannot be mixed with other "
```
davisp: You need to remove the ", not" as well.

src/couch_query_servers.erl
```diff
@@ -169,8 +177,20 @@ sum_values(Value, Acc) when is_number(Value), is_list(Acc) ->
     sum_arrays(Acc, [Value]);
 sum_values(Value, Acc) when is_list(Value), is_number(Acc) ->
     sum_arrays([Acc], Value);
-sum_values(Else, _Acc) ->
-    throw_sum_error(Else).
```
davisp: This clause needs to be reinstated and fixed to return the proper error object.

src/couch_query_servers.erl
```diff
@@ -489,8 +509,8 @@ ret_os_process(Proc) ->
     catch unlink(Proc#proc.pid),
     ok.
 
-throw_sum_error(Else) ->
-    throw({invalid_value, iolist_to_binary(io_lib:format(?SUMERROR, [Else]))}).
+throw_sum_error(_Else) ->
+    throw({invalid_value, ?SUMERROR}).
```
davisp: You undid the guts of the patch?

davisp:
That was a really confusing description in hindsight...

Consider it this way: what you're changing could easily be written as a custom reduce function in JavaScript. The only difference between the two is that one is in Erlang and there's some goop about aliasing the implementation with a "_name". That's it.

And given that we know that reduce functions can return whatever they want we know that there's no reason that we have to change anything related to chttpd or fabric or invent changes to the worker protocol that's sending messages around.

src/couch_query_servers.erl
tonysun83: eh not sure how that one got left out

davisp:
My bad, I missed the fact that we were talking about removing the top level value. I don't think we should do that in this case because of what I described earlier. These are just alternative implementations of reduce functions. As such they can return whatever we want them to although it should probably stay within the confines of the existing protocol.

As to the point that users could write a reduce function that maps to the same error object, that's precisely the point. Users could write this themselves in JavaScript. The only reason that builtins exist is for performance for the common cases of counting, summing, and statistics. Inventing a new protocol for builtin reduces seems a bit of a crossing of abstractions. Granted we could go and create a whole new protocol that custom reduce functions could use, but that seems significantly different than the original intention of this ticket.

For extra specificity the full response would look something like such:

{
  "rows": [
    {"key": $whatever, "value":{"error": "builtin_reduce_error", "reason": "stuff", "caused_by": bad_value}},
    ...
]}

If we want to change the row definition there then we do have to go and start changing our protocols which does require things like multiple rolling reboots. However if we do go that direction we should think harder and make it an available option for custom reduce functions as well.

tonysun83:
Moved the messages to the lower level again, so the new and hopefully final format:

```
{
    "rows": [
        {
            "key": null,
            "value": {
                "error": "builtin_reduce_error",
                "reason": "The _sum function requires that map values be numbers, arrays of numbers, or objects. Objects cannot be mixed with other data structures. Objects can be arbitrarily nested, provided that the values for all fields are themselves numbers, arrays of numbers, or objects.",
                "caused_by": [
                    {
                        "a": 1
                    },
                    {
                        "a": 2
                    },
                    {
                        "a": 3
                    },
                    {
                        "a": 4
                    }
                ]
            }
        }
    ]
}
```

kxepal:
@tonysun83 
Nice work! One more little question: how much number of items will be in `caused_by` list? Everything what map function produced to reduce or just those ones which had caused an issue?

tonysun83:
@kxepal : it will be the first one that caused the issue. From my tests if doc1, doc2 were valid and doc3 was invalid, then only the input from doc3 was shown. However, if doc1, doc2, and doc3 were all invalid, it would sometimes show doc1, sometimes doc2, and sometimes doc3. I think that's expected because of how the reducer loads the input from the btree.

kxepal:
@tonysun83 
Good, thanks . I worried about to get there all the map data or some else large batch which could cause a troubles to receive, parse, load and process. Glad that this won't happen (:

src/couch_query_servers.erl
```diff
@@ -142,25 +142,34 @@ os_rereduce(Lang, OsRedSrcs, KVs) ->
...
+    Acc;
+builtin_sum_rows([[Key, Value] | RestKVs], Acc) ->
+    try sum_values(Value, Acc) of
+        NewAcc ->
+            builtin_sum_rows(RestKVs, NewAcc)
+    catch
+        throw:{builtin_reduce_error, Obj} ->
+            Obj;
+        throw:{invalid_value, Reason, Cause} ->
+            {[{<<"error">>, <<"builtin_reduce_error">>},
```
iilyak: Why we have to handle the same error twice. Am I understanding it correctly? 
1. we catch invalid_value
2. we return `{[{<<"error">>, ...}]}`
3. on the outer recursion call we go into sum_values
4. sum_values detects the `{[{<<"error">>, ...}]}` and throws `{builtin_reduce_error,...`
5. we catch `{builtin_reduce_error,...` and return Obj which is our `{[{<<"error">>, ...}]}`

Sounds complicated.

davisp:
Reduce functions are complicated so that makes sense. :D

Unfortunately there's not a whole lot we can do without inventing a builtin-reduce specific error API/protocol which strikes me as a bit ugly considering it leaves non-builtin-reduce functions out. I'll make comments for each of your notes individually.

src/couch_query_servers.erl
```diff
@@ -142,25 +142,34 @@ os_rereduce(Lang, OsRedSrcs, KVs) ->
...
-builtin_sum_rows(KVs) ->
-    lists:foldl(fun([_Key, Value], Acc) -> sum_values(Value, Acc) end, 0, KVs).
 
-sum_values({Props}, 0) ->
-    {Props};
-sum_values({Props}, {AccProps}) ->
-    {sum_objects(lists:sort(Props), lists:sort(AccProps))};
+builtin_sum_rows([], Acc) ->
+    Acc;
+builtin_sum_rows([[Key, Value] | RestKVs], Acc) ->
```
davisp: Good catch.

src/couch_query_servers.erl
davisp: This is because a reduction function has to operate on its own output, not because of recursion. What we're doing here is setting up a try/catch so that we can have a non-local return when we detect an error. That error might either be when doing a reduce and hitting a bad value emitted in the index, or it might be later on when we're rereducing a group of reductions that contains an error. In both cases we want to pass the error up immediately and cease recursion.

Technically we could have split this into two different functions that are called based on reduce vs rereduce but the existing logic precluded that from being very simple without a massively more complicated change.

src/couch_query_servers.erl
```diff
@@ -169,6 +178,19 @@ sum_values(Value, Acc) when is_number(Value), is_list(Acc) ->
     sum_arrays(Acc, [Value]);
 sum_values(Value, Acc) when is_list(Value), is_number(Acc) ->
     sum_arrays([Acc], Value);
+sum_values({Props}, Acc) ->
+    case lists:keyfind(<<"error">>, 1, Props) of
```
davisp: Also unfortunately but we can't return anything but an EJSON value from this because there's no way to insert a convert-to-EJSON step later on. And any attempt to do so would be outside of the standard reduce function APIs and would prevent any non-builtin reduce from participating in that protocol (cause other language can only pass JSON around).

tonysun83:
@iilyak : changed unused ref

iilyak:
+1. All tests pass locally
```
    Running test function(s):
      couch_query_servers:builtin_sum_rows_negative_test/0
      couch_query_servers:sum_values_negative_test/0
======================== EUnit ========================
couch_query_servers: sum_values_negative_test...[0.002 s] ok
couch_query_servers: builtin_sum_rows_negative_test...ok
=======================================================
  All 2 tests passed.
```
```
==> couch (eunit)
...
=======================================================
  All 792 tests passed.
```

PR #230, opened 2017-02-22
Add sys db callbacks only once per db open

brkolla:
Currently we are adding the sys db callbacks for every request instead once
while opening the db. This fix will ensure that we will only add the sys
db calls backs once while opening the db.

COUCHDB-3307

### DIFF
--- a/src/couch_server.erl
+++ b/src/couch_server.erl
@@ -71,18 +71,17 @@ sup_start_link() ->
 
 
 open(DbName, Options0) ->
-    Options = maybe_add_sys_db_callbacks(DbName, Options0),
-    Ctx = couch_util:get_value(user_ctx, Options, #user_ctx{}),
+    Ctx = couch_util:get_value(user_ctx, Options0, #user_ctx{}),
     case ets:lookup(couch_dbs, DbName) of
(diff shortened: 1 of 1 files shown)


nickva:
It seems functionally there is a difference between the current code path and the new one:

If we don't add callbacks in case db is found, we don't update `Options` to include `sys_db` marker, so then then when we call `update_lru` it won't be inserted it in the LRU. Did we mean to do that?  At least we should mention the side-effect in the commit.

Maybe we want to grab Options from `[#db{fd=Fd, fd_monitor=Lock} = Db] when Lock =/= locked ->` line instead of Options0 then re-run benchmark. Wonder if we are seeing a performance improvement not just directly from `maybe_add_sys_db_callbacks` function not taking more time (it does 3 config reads), but from the side effect of not updating LRU on every open.

brkolla:
@nickva Good catch. Will use the _Options_ from **db** record and rerun the test.

davisp:
Looks good. I'd also go ahead and remove the update_lru call on line 85 as its superflous. If we just got a message back from couch_server that means it was just placed into the LRU so updating it again is a double/worthless update to the ordering.

brkolla:
@davisp Made the change as per your suggestion. Please review and approve and merge once the CI checks pass. Not sure why the CI check is in pending state forever.

PR #231, opened 2017-03-03
Fix `badarith` error in couch_db:get_db_info call

nickva:
When folding we account for a previous `null`, `undefined`, or a number. However
btree:size/1 returns 0, `nil` or a number.

Switched `undefined` to `nil`. Couldn't find where btree:size would ever return
`undefined`, it seems we meant to use `nil` instead.

COUCHDB-3316


rnewson:
+1

davisp:
+1

PR #233, opened 2017-03-07
Rename max_document_size to max_http_request_size

nickva:
`max_document_size` is implemented as `max_http_request_size`. There was no
real check for document size. In some cases the implementation was close enough
of a proxy (PUT-ing and GET-ing single docs), but in some edge cases, like
_bulk_docs requests the discrepancy between request size and document size
could be rather large.

The section was changed accordingly from `couchdb` to `httpd`. `httpd` was
chosen as it applies to both clustered as well as local interface.

There is a parallel effort to implement an actual max_document_size check. The
set of commit should be merged close enough together to allow for a backwards
compatible transition.

COUCHDB-2992

### DIFF
--- a/src/couch_httpd.erl
+++ b/src/couch_httpd.erl
@@ -451,7 +451,7 @@ validate_ctype(Req, Ctype) ->
 
 check_max_request_length(Req) ->
     Len = list_to_integer(header_value(Req, "Content-Length", "0")),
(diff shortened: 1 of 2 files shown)


nickva:
Related PRs:

 * https://github.com/apache/couchdb-couch-replicator/pull/61
 * https://github.com/apache/couchdb-chttpd/pull/156

PR #235, opened 2017-03-13
Allow limiting maximum document body size

nickva:
Configuration is via the `couchdb.max_document_size`. In the past that
was implemented as a maximum http request body size and this finally implements
it by actually checking a document's body size.

COUCHDB-2992

### DIFF
--- a/src/couch_doc.erl
+++ b/src/couch_doc.erl
@@ -13,7 +13,7 @@
 -module(couch_doc).
 
 -export([to_doc_info/1,to_doc_info_path/1,parse_rev/1,parse_revs/1,rev_to_str/1,revs_to_strs/1]).
--export([from_json_obj/1,to_json_obj/2,has_stubs/1, merge_stubs/2]).
+-export([from_json_obj/1, from_json_obj_validate/1, to_json_obj/2,has_stubs/1, merge_stubs/2]).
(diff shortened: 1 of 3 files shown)


src/couch_doc.erl
```diff
@@ -125,7 +125,14 @@ doc_to_json_obj(#doc{id=Id,deleted=Del,body=Body,revs={Start, RevIds},
     }.
 
 from_json_obj({Props}) ->
```
sagelywizard: Is this the right place to be making this change? `from_json_obj` is used in places other than when parsing HTTP request bodies. For example, this change will cause `fabric:cleanup_index_files` to fail if any existing design documents are larger than the limit.

nickva: Hmm good point. It seems like the cleanest place code-wise as we do other validations in that function (`throw({doc_validation,...`). Wonder if it makes sens to make a validating and non-validation version of the function.

nickva:
@sagelywizard I added a validating version of the function. The other code like cleanup_index_files can still call the old function and not crash on size limit change.

nickva:
Related pr:

 * https://github.com/apache/couchdb-fabric/pull/91

 * https://github.com/apache/couchdb-chttpd/pull/157

nickva:
@sagelywizard should be a bit better now. But ended up with another pr for fabric as well.

PR #237, opened 2017-03-14
Add sys_dbs to the LRU

rnewson:
_replicator and _users db's have been excluded from the LRU for a long time. This is fine when you have one of each, it's not fine when you have tens of thousands of each.

Note: I've intentionally changed max_dbs_open to be a soft limit rather than a hard one. couchdb's behavior when it throws all_dbs_active is pretty poor (you can create a database that you can't open later, for example) so I think it better to remove that class of problems.

COUCHDB-3325

### DIFF
--- a/src/couch_server.erl
+++ b/src/couch_server.erl
@@ -73,8 +73,8 @@ sup_start_link() ->
 open(DbName, Options0) ->
     Ctx = couch_util:get_value(user_ctx, Options0, #user_ctx{}),
     case ets:lookup(couch_dbs, DbName) of
-    [#db{fd=Fd, fd_monitor=Lock, options=Options} = Db] when Lock =/= locked ->
-        update_lru(DbName, Options),
+    [#db{fd=Fd, fd_monitor=Lock} = Db] when Lock =/= locked ->
+        update_lru(DbName),
         {ok, Db#db{user_ctx=Ctx, fd_monitor=erlang:monitor(process,Fd)}};
     _ ->
         Options = maybe_add_sys_db_callbacks(DbName, Options0),
@@ -91,11 +91,8 @@ open(DbName, Options0) ->
         end
     end.
 
-update_lru(DbName, Options) ->
-    case lists:member(sys_db, Options) of
-        false -> gen_server:cast(couch_server, {update_lru, DbName});
-        true -> ok
-    end.
+update_lru(DbName) ->
+    gen_server:cast(couch_server, {update_lru, DbName}).
 
 close_lru() ->
     gen_server:call(couch_server, close_lru).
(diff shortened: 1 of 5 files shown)


src/couch_server.erl
```diff
@@ -91,11 +91,8 @@ open(DbName, Options0) ->
         end
     end.
 
-update_lru(DbName, Options) ->
-    case lists:member(sys_db, Options) of
-        false -> gen_server:cast(couch_server, {update_lru, DbName});
-        true -> ok
-    end.
+update_lru(DbName, _Options) ->
```
sagelywizard: Can we change this to `update_lru/1`?

src/couch_server.erl
```diff
@@ -271,20 +268,18 @@ all_databases(Fun, Acc0) ->
     {ok, FinalAcc}.
 
 
-make_room(Server, Options) ->
-    case lists:member(sys_db, Options) of
-        false -> maybe_close_lru_db(Server);
-        true -> {ok, Server}
-    end.
+make_room(Server, _Options) ->
```
sagelywizard: Can we get rid of this function and just use `maybe_close_lru_db/1`?

src/couch_server.erl
```diff
@@ -542,17 +530,11 @@ handle_info(restart_config_listener, State) ->
 handle_info(Info, Server) ->
     {stop, {unknown_message, Info}, Server}.
 
-db_opened(Server, Options) ->
-    case lists:member(sys_db, Options) of
-        false -> Server#server{dbs_open=Server#server.dbs_open + 1};
-        true -> Server
-    end.
+db_opened(Server, _Options) ->
```
sagelywizard: Same with `db_opened/2` and `db_closed/2`. Can we just change the arity?

src/couch_server.erl
rnewson: I was leaving sigs intact in case hot upgrade

eiri:
Looks good to me.

For a reference [here is what I came up with](https://github.com/apache/couchdb-couch/compare/master...cloudant:add-lru-excluded-option)

sagelywizard:
I think test/couchdb_view_tests.erl needs to be updated.

src/couch_server.erl
sagelywizard: Though it'll only be problematic if you force an upgrade mid-call to `couch_db:open/2`, right? Might be worth leaving in just so those don't crash, but it'd be nice to have a path going forward to remove that cruft.

src/couch_server.erl
sagelywizard: Since `make_room/2` is only called by couch_server, can't we safely remove this?

src/couch_server.erl
sagelywizard: See: `make_room`

rnewson:
ah, true, yes.

nickva:
Tried it locally. Looks good. 

+1 After tests are fixed.

We'd want to performance tests this at some point.

rnewson:
tests fixed (sorry @eiri I got there first)

PR #241, opened 2017-03-30
Improve compaction task status updates

davisp:
Previous the emsort related operations did not update the compaction
task status. For large databases this leads to some very long waits
while the compaction task stays at 100%. This change adds progress
reports to the steps for sorting and copying document ids back into the
database file.

### DIFF
--- a/src/couch_db_updater.erl
+++ b/src/couch_db_updater.erl
@@ -34,6 +34,8 @@
     infos
 }).
 
+-define(COMP_DOCID_BATCH_SIZE, 1000).
+
 init({DbName, Filepath, Fd, Options}) ->
     erlang:put(io_priority, {db_update, DbName}),
     case lists:member(create, Options) of
@@ -1108,10 +1110,10 @@ copy_docs(Db, #db{fd = DestFd} = NewDb, MixedInfos, Retry) ->
     NewDb#db{id_tree=IdEms, seq_tree=SeqTree}.
 
 
-copy_compact(Db, NewDb0, Retry) ->
+copy_compact(Db, NewDb0, Retry, TotalChanges) ->
     Compression = couch_compress:get_compression_method(),
     NewDb = NewDb0#db{compression=Compression},
-    TotalChanges = couch_db:count_changes_since(Db, NewDb#db.update_seq),
+
     BufferSize = list_to_integer(
         config:get("database_compaction", "doc_buffer_size", "524288")),
     CheckpointAfter = couch_util:to_integer(
@@ -1147,6 +1149,7 @@ copy_compact(Db, NewDb0, Retry) ->
     TaskProps0 = [
         {type, database_compaction},
         {database, Db#db.name},
+        {phase, seq_tree},
         {progress, 0},
         {changes_done, 0},
         {total_changes, TotalChanges}
@@ -1193,6 +1196,8 @@ start_copy_compact(#db{}=Db) ->
         open_compaction_files(Name, Header, Filepath, Options),
     erlang:monitor(process, MFd),
 
+    TotalChanges = couch_db:count_changes_since(Db, NewDb#db.update_seq),
+
     % This is a bit worrisome. init_db/4 will monitor the data fd
     % but it doesn't know about the meta fd. For now I'll maintain
     % that the data fd is the old normal fd and meta fd is special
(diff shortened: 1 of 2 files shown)


src/couch_db_updater.erl
```diff
@@ -1323,12 +1326,82 @@ bind_id_tree(Db, Fd, State) ->
...
+        {event_st, {init, 0, 0}}
+    ],
+    Ems1 = couch_emsort:set_options(Ems0, Options),
+    {ok, Ems2} = couch_emsort:merge(Ems1),
+    Db0#db{id_tree=Ems2}.
+
+
+emsort_cb(_Ems, {merge, chain}, {init, Copied, Nodes}) ->
+    {init, Copied, Nodes + 1};
+emsort_cb(_Ems, row_copy, {init, Copied, Nodes}) when Copied >= 1000 ->
```
iilyak: Magic constant 1000.

src/couch_db_updater.erl
```diff
@@ -1323,12 +1326,82 @@ bind_id_tree(Db, Fd, State) ->
...
+
+    couch_task_status:update([
+        {phase, sort_ids},
+        {total_changes, Iters * TotalCopied},
+        {changes_done, 0},
+        {progress, 0}
+    ]),
+    0;
+
+emsort_cb(_Ems, row_copy, Copied) when is_integer(Copied), Copied > 1000 ->
```
iilyak: Magic constant 1000.

src/couch_db_updater.erl
```diff
@@ -1323,12 +1326,82 @@ bind_id_tree(Db, Fd, State) ->
...
+
+
+emsort_cb(_Ems, {merge, chain}, {init, Copied, Nodes}) ->
+    {init, Copied, Nodes + 1};
+emsort_cb(_Ems, row_copy, {init, Copied, Nodes}) when Copied >= 1000 ->
+    update_compact_task(Copied + 1),
+    {init, 0, Nodes};
+emsort_cb(_Ems, row_copy, {init, Copied, Nodes}) ->
+    {init, Copied + 1, Nodes};
+emsort_cb(Ems, {merge_start, reverse}, {init, Copied, Nodes}) ->
```
iilyak, line 1356: I noticed that we can also have `{merge_start, forward}` [event](https://github.com/apache/couchdb-couch/blob/1db1337301a7c897be41e13cd328270509008478/src/couch_emsort.erl#L246:L252). Should we handle it as well?

src/couch_db_updater.erl
davisp: I assume you mean you'd rather that be a define? I just copied the same shape as merge_docids down below.

davisp:
Also I need to change my use of DocCount to be TotalChanges. I noticed while testing the optimizations I tried that its currently not correct during compaction retries.

src/couch_db_updater.erl
iilyak: It looks like we wouldn't have a progress updates for the second pass. Is it intentional?

iilyak (commented):
It looks good but I have few questions.

src/couch_db_updater.erl
iilyak: Yes I mean that it would be better to have a define. Like ?UPDATE_FREQ or ?BATCH_SIZE or something.

davisp:
Updated to use define's for batch sizes and fixed the changes/doc count mixup.

For the {merge_start, forward}, thats only there for completeness. It felt a bit weird to not include that event even though I don't necessarily need it for this work.

Updates during the second phase happen here:

https://github.com/apache/couchdb-couch/pull/241/files#diff-f6f654ab26b490bab95be6f502c49d89R1376

This is a bit subtle but it seemed like the best I could do without starting to modify the on-disk contents of files which I like to avoid when at all possible.

Its a bit funky but without starting to store the total number of rows in emsort the best approach I had was to guess with the input being the total number of changes processed in this run and then in the first phase of the merge sort just count how many rows we have. Once that first pass is over we can calculate the total number of rows that will be copied and then just update progress in the normal fashion. Its a bit awkward but that emsort_cb kind of switches modes once it sees that the first phase has ended and then its purely just listening for row_copy events.

davisp:
And I hesitate to store the total number of rows because there'd then be upgrade things to worry about and generally speaking this will just sort itself out while still giving an approximate progress update.

